PR #2, opened 2021-10-02
Omar/ci backend

obaranek:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Configure-CI-backend-633e3b3182764d3d9deeead74da67751)

## Implementation description
* fixed linting errors in starter code
* added ci-backend workflow

## Steps to test
1. Run yarn lint and verify that there are no errors
2. trigger github actions and verify all steps run successfully 

* 

## Checklist

### DIFF
--- a/.github/workflows/ci-backend.yml
+++ b/.github/workflows/ci-backend.yml
@@ -1,0 +1,35 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI_BACKEND
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the development branch
+  push:
+    branches: [ development ]
+  pull_request:
+    branches: [ development ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+      
+      # Install dependencies
+      - name: Install dependencies
+        run:  yarn --cwd ./backend/typescript/
+        
+      #Lint Check
+      - name: Run lint for backend
+        working-directory: ./backend/typescript
+        run: yarn lint
--- a/backend/typescript/models/index.ts
+++ b/backend/typescript/models/index.ts
@@ -1,7 +1,7 @@
 import * as path from "path";
 import { Sequelize } from "sequelize-typescript";
 
-export const sequelize = new Sequelize(
+export default sequelize = new Sequelize(
   `postgres://${process.env.POSTGRES_USER}:${process.env.POSTGRES_PASSWORD}@${process.env.DB_HOST}:5432/${process.env.POSTGRES_DB}`,
   { models: [path.join(__dirname, "/*.model.ts")] },
 );
--- a/backend/typescript/migrate.js
+++ b/backend/typescript/migrate.js
@@ -1,3 +1,5 @@
+import { migrator } from "./umzug";
+
 require("ts-node/register");
 
-require("./umzug").migrator.runAsCLI();
+migrator.runAsCLI();
--- a/backend/typescript/models/entity.model.ts
+++ b/backend/typescript/models/entity.model.ts
@@ -22,4 +22,3 @@ export default class Entity extends Model {
   @Column
   file_name!: string;
 }
-
--- a/backend/typescript/rest/userRoutes.ts
+++ b/backend/typescript/rest/userRoutes.ts
@@ -94,7 +94,7 @@ userRouter.post("/", createUserDtoValidator, async (req, res) => {
     });
 
     await authService.sendEmailVerificationLink(req.body.email);
-    
+
     res.status(201).json(newUser);
   } catch (error) {
     res.status(500).json({ error: error.message });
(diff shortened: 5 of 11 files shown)


Tayef-Shah:
can you quickly rename the PR title to the specific change? Something like "add CI config for backend"

.github/workflows/ci-backend.yml
```diff
@@ -0,0 +1,35 @@
...
+on:
+  # Triggers the workflow on push or pull request events but only for the development branch
+  push:
+    branches: [ development ]
+  pull_request:
+    branches: [ development ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
```
Tayef-Shah, line 15: is the value supposed to be blank or should it say false or something?

.github/workflows/ci-backend.yml
```diff
@@ -0,0 +1,35 @@
...
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
```
Tayef-Shah, line 21: should we always do latest, or should we specify a specific version that matches with our backend hosting?

Tayef-Shah (commented):
looks good overall, just had a couple of questions

.github/workflows/ci-backend.yml
obaranek: I think usually there would be options to add under worflow_dispatch, but for now, I think we just need to add the manual trigger.
https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

.github/workflows/ci-backend.yml
obaranek: Good question, @Tahmeed27 what do you think?

obaranek: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners. This is a list of the host runners github provides.

.github/workflows/ci-backend.yml
Tayef-Shah: having a tag for each manual run would be helpful to know what the purpose was. So maybe we can add a key like this to the workflow_dispatch
```
tags:
        description: 'Test scenario tags' 
```

obaranek: Sounds good, I'll add that.

.github/workflows/ci-backend.yml
Tayef-Shah: I'm down to make it latest/ubuntu-20.04 for now, and have a ticket to update it when we have deployments running

Tayef-Shah (approved):
LGTM. Just remember to add a ticket for updating the ubuntu version to match the backend deployment.

.github/workflows/ci-backend.yml
Tahmeed27: Yeah I say keep it latest for now

.github/workflows/ci-backend.yml
```diff
@@ -0,0 +1,38 @@
...
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+    inputs:
+      tags:
+        description: 'Test scenario tags'
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
```
Tahmeed27, line 22: Name the job backend_lint and update the comment

.github/workflows/ci-backend.yml
```diff
@@ -0,0 +1,38 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI_BACKEND
```
Tahmeed27, line 3: Name this: Perform Lint Checks on Backend

.github/workflows/ci-backend.yml
```diff
@@ -0,0 +1,38 @@
...
+  push:
+    branches: [ development ]
+  pull_request:
+    branches: [ development ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+    inputs:
+      tags:
+        description: 'Test scenario tags'
```
Tahmeed27, line 17: Why does this say test scenario tags? Is it from wherever you copy pasted the code? Seems out of place

.github/workflows/ci-backend.yml
```diff
@@ -0,0 +1,38 @@
...
+      - uses: actions/checkout@v2
+      
+      # Install dependencies
+      - name: Install dependencies
+        run:  yarn --cwd ./backend/typescript/
+        
+      #Lint Check
+      - name: Run lint for backend
+        working-directory: ./backend/typescript
+        run: yarn lint
```
Tahmeed27, line 38: Can you set it so that the lint check fails if there are errors present? Also how bad will it be to also get the code to a point where there are no warnings either?

backend/typescript/models/index.ts
```diff
@@ -1,7 +1,7 @@
 import * as path from "path";
 import { Sequelize } from "sequelize-typescript";
 
-export const sequelize = new Sequelize(
+export default sequelize = new Sequelize(
```
Tahmeed27, line 4: out of curiosity  - when do you use export default vs const? Which es-lint rule was this

Tahmeed27 (changes requested):

backend/typescript/models/index.ts
obaranek: it's named export vs default export. 
Since I think we're exporting one thing in the file, the linting recommend using default export so that you can just import using the import default syntax: `import name_of_reference from 'index.ts'`
as opposed to named import which you have to specify the name of what you're importing: `import {sequelize} from 'index.ts`

.github/workflows/ci-backend.yml
obaranek: GitHub actions error out automatically when the exit code is 1. [https://github.com/uwblueprint/community-fridge-kw/actions/runs/1273715914](This is a link to community fridge CI workflow) where the job fails on the linting error.

PR #16, opened 2021-11-07
able to create user with rand password and send login details via email

dhmi19:
## Notion ticket link

[Admin Authentication: Create Account] (https://www.notion.so/uwblueprintexecs/Admin-Authentication-Create-Account-b2b23688973543458dc621ada42fc469)

## Implementation description
* made a new function called sendPasswordSetupLink() that sends an email to the user with their temp login details. Decided to make a new function as the name was more suitable and the functionality was tailored to giving temp details and having a link to set up password. Also, this should not affect changes elsewhere as it is a new function.

## Steps to test
1. Create a new user from the front end with first name, email, password
2. Verify that user account was created in firebase and sql database
3. verify that you got an email with your login details and a random password 

* should only be minor changes, focus mostly on the new function sendPasswordSetupLink() and the ./register route. 

## Checklist


backend/typescript/yarn.lock
```diff
@@ -4,26 +4,26 @@
 
 "@babel/code-frame@7.12.11":
   version "7.12.11"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.12.11.tgz#f4ad435aa263db935b8f10f2c552d23fb716a63f"
```
dhmi19, line 7: Not sure why these changes are made in yarn.lock. 

Could someone verify if this is ok? @Tahmeed27 @Tayef-Shah

Tahmeed27 (changes requested):
Looks good other than the couple minor changes. One thing I noticed whilst testing is this - When you login as a new user and then try to do any thing from the Default page such as log out or reset password, none of it actually works. And that's because they don't let you do any of these tasks whilst you're unverified.So while this is okay in this ticket since our verification will be through the setupPassword() handler, if you're testing users and nothing is responding, this is probably why. You can add in the sendEmailVerification link to something like login to get your account verified to test those features again.

Also another note - the email is still going to my spam account. See if you can search that up and figure out if there's any way to not flag our email service so it can be treated like a normal email. Changes for that doesn't have to be a part of this ticket

backend/typescript/yarn.lock
Tahmeed27: Hmm its interesting that it changed to npm instead of yarn. Idk if that's because you did npm install at some point or not tho. You should only use yarn for dependencies and not mix it up with npm. Maybe you could delete yarn.lock and then run yarn again to see if that changes this file?

Tahmeed27 (approved):
LGTM

PR #20, opened 2021-11-13
add seeders and update README

Tayef-Shah:
## Notion ticket link

[Seed Database](https://www.notion.so/uwblueprintexecs/Seed-Database-a0b0e500155e4f63a678f20c635db14f)

## Implementation description
* Add seed script and seed migrations
* Seed migrations added with the command `node seed create --name my-migration.js`

## Steps to test
1. start containers: `docker-compose up -d`
2. run bash in backend container `docker-compose exec ts-backend bash`
3. make another terminal tab and run bash in db container and log in
```
docker-compose exec db bash
psql -U postgres ccbc
```
4. run `node seed up` in the backend container to seed the db
5. inspect the database has been seeded with 5 reviews and related data (including associations)
6.  run `node seed down --to 0` to delete all the seed data
7. Inspect that it has been deleted

* reliability (works after multiple invocations of the commands)

## Checklist


Tahmeed27 (changes requested):
So I was trying to test it and two things:
- Since I already had some test review data in tables, I realized node seed up fails because the id matches. So do you think there's a way to drop all tables first and then seed the data or do you think maybe the seed data should be using some super high id to make it work with the already pre-existing data?
- So after I did delete all the review data and ran node seed up, my table seems to be populated but I think it might not be populating correctly because I ran get Reviews afterwards and this is the error that I got. Could you look into this too? 
image

Tayef-Shah:
@Tahmeed27 I pushed a fix, feel free to test it out

Tahmeed27 (approved):
Great job! Looks like it's working to me :D

PR #24, opened 2021-11-22
Change test db url to correct version

Tahmeed27:
## Implementation description
* Altered test db url to use the test db rather than main db


Tayef-Shah (approved):
LGTM

PR #25, opened 2021-11-25
Dhruv/custom admin signup

dhmi19:
## Notion ticket link

[Admin Authentication: Extend Signup Time](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d)

## Implementation description
Added custom email handler url which is generated in the AuthService
Added component + route to front end for password resets, setting up passwords, resetting email etc. 
Added a firebase client as a singleton on the frontend (inside Firebase.ts file) so we can use firebase functions on the front end that need the auth object (check to see if this is the best option)

## Steps to test
1. Have 2 windows open (1 for signing up a user and the other to open the email and set the password)
2. Make sure both windows have cleared local storage
3. Signup a user on window 1 by entering their first name, second name, email in url: localhost:3000/signup
4. On window 2, open the email and try to change the password 
5. You should be verified and the password should be changed. 
6. After you have set up your password, you should not be able to go to the set up password page again (to check this, clear the local storage then try to access that link from your email again)

* Focus on checking if I missed any part of the implementation as it  is a multistep process
* There is an error which gets thrown on the frontend when we try to set the password after the user clicks the link in their email
* Check to see if the firebase singleton and the getAuth() functionality in file SetPassword.tsx is the best way to go

## Checklist

### DIFF
--- a/backend/typescript/rest/authRoutes.ts
+++ b/backend/typescript/rest/authRoutes.ts
@@ -43,9 +43,19 @@ authRouter.post("/login", loginRequestValidator, async (req, res) => {
   }
 });
 
+/* returns a firebase user given a user id */
+authRouter.get("/:uid", async (req, res) => {
+  try {
+    const firebaseUser = await authService.getFirebaseUserByUid(req.params.uid);
+    res.status(200).json(firebaseUser);
+  } catch (error) {
+    sendErrorResponse(error, res);
+  }
+});
+
 /* Register a user, returns access token and user info in response body and sets refreshToken as an httpOnly cookie */
 authRouter.post("/register", registerRequestValidator, async (req, res) => {
-  const randomPassword = password.randomPassword({
+  const accessCode = password.randomPassword({
     length: 8, // length of the password
     characters: [
       // acceptable characters in the password
@@ -63,20 +73,17 @@ authRouter.post("/register", registerRequestValidator, async (req, res) => {
       lastName: req.body.lastName,
       email: req.body.email,
       role: "Admin", // TODO: pass in the role as a parameter to function for author and subscriber
-      password: randomPassword,
+      password: accessCode,
     });
 
     // try to sign in the user and return the expiring token
-    const authDTO = await authService.generateToken(
-      req.body.email,
-      randomPassword,
-    );
+    const authDTO = await authService.generateToken(req.body.email, accessCode);
 
     const { refreshToken, ...rest } = authDTO;
 
     // Send email with login details and ask to change password
-    // once they change the password, admin should be verified
-    await authService.sendPasswordSetupLink(req.body.email, createdUser);
+    // once they change the password, user should be verified
+    await authService.sendPasswordSetupLink(createdUser, accessCode);
 
     res
       .cookie("refreshToken", refreshToken, {
--- a/backend/typescript/rest/authRoutes.ts
+++ b/backend/typescript/rest/authRoutes.ts
@@ -153,7 +153,7 @@ authRouter.post(
 authRouter.post("/verifyEmail/:uid", async (req, res) => {
   try {
     await authService.verifyEmail(req.params.uid);
-    res.status(200).send();
+    res.status(204).send();
   } catch (error: unknown) {
     sendErrorResponse(error, res);
   }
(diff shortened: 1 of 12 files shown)


backend/typescript/rest/authRoutes.ts
```diff
@@ -43,9 +43,19 @@ authRouter.post("/login", loginRequestValidator, async (req, res) => {
   }
 });
 
+/* returns a firebase user given a user id */
+authRouter.get("/:uid", async (req, res) => {
```
Tahmeed27, line 47: Did you miss adding this end-point to the swagger.yml?

frontend/src/APIClients/AuthAPIClient.ts
```diff
@@ -126,12 +111,49 @@ const refresh = async (): Promise<boolean> => {
...
 
+/**
+ * Retrieve the firebase user given a user id
+ * @param uid user's id
+ * @returns User: a firebase user instance
+ * @throws error if could not retrieve user by uid
+ */
+const getFirebaseUserByUid = async (uid: string): Promise<User> => {
+  const { data } = await baseAPIClient.get(`/auth/${uid}`, {
+    withCredentials: true,
```
Tahmeed27, line 122: What is this withCredentials: true for?

frontend/src/APIClients/AuthAPIClient.ts
```diff
@@ -126,12 +111,49 @@ const refresh = async (): Promise<boolean> => {
...
+const verifyEmail = async (uid: string): Promise<boolean> => {
+  const bearerToken = `Bearer ${getLocalStorageObjProperty(
+    AUTHENTICATED_USER_KEY,
+    "accessToken",
+  )}`;
+  try {
+    await baseAPIClient.post(
+      `/auth/verifyEmail/${uid}`,
+      {},
+      { headers: { Authorization: bearerToken } },
```
Tahmeed27, line 142: Hmm do you still need to pass bearerToken since you don't have any middleware for authorization?

frontend/src/utils/Firebase.ts
```diff
@@ -0,0 +1,14 @@
+import * as firebase from "firebase/app";
```
Tahmeed27, line 1: Minor nit but can u rename this to firebase.ts

Tahmeed27 (changes requested):
Great work and thanks for the super detailed PR description and documentation

frontend/src/utils/Firebase.ts
dhmi19: renamed file to firebase.ts

backend/typescript/rest/authRoutes.ts
dhmi19: @Tahmeed27 do we need to protect this endpoint? We need it so that unauthorized users and even unregistered users can get the firebaseUser but as of now, anyone can call this endpoint to get our user right?

backend/typescript/services/implementations/authService.ts
```diff
@@ -177,26 +180,27 @@ class AuthService implements IAuthService {
...
     try {
-      const passwordResetLink = await firebaseAdmin
-        .auth()
-        .generatePasswordResetLink(email);
-      // first-time determines if we are setting a new account password
-      // (account will be verified) or reseting an old account's password
-      const setPasswordLink = passwordResetLink.concat("&first-time=true");
+      const authId = await this.userService.getAuthIdById(user.id);
+
+      const setPasswordLink = `http://localhost:3000/auth/action?mode=setup-password&uid=${authId}`;
```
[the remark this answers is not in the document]
Tayef-Shah: We don't have anything exactly like that (with the port), but we have `FIREBASE_URI`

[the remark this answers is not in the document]
dhmi19: I think we'll need one for local host right? this is to navigate to a front end page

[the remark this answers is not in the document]
Tayef-Shah: Let's make an env var called CLIENT_URL

Tahmeed27 (changes requested):
Also please look into why the ci tests are failing all of them should pass now

backend/typescript/rest/authRoutes.ts
Tayef-Shah: Does it expose anything sensitive? If it's just Firebase User, then it should be fine right?

dhmi19:
Closing pull request because I split the authentication flow into 2 pages. Made a new PR which builds on top of this but splits the workflow into 2 steps

PR #26, opened 2021-11-27
Omar/create review service

obaranek:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Create-Review-Service-4e0fa6707a164d118aad9f914fea42a3)

## Implementation description
* added the createReviewService
* added the endpoint for REST routes
* added 2 unit tests in jest
* added swagger API documentation 

## Steps to test
1. run the test command in the ts-backend container

* 

## Checklist


backend/typescript/services/implementations/ReviewService.ts
```diff
@@ -0,0 +1,174 @@
...
+        );
+
+        const pgTagsRet: PgTag[] = await Promise.all(
+          review.tags.map(async (reviewTag) => {
+            const tag = await PgTag.findOrCreate({
+              where: { name: reviewTag.name },
+              transaction: t,
+            }).then((data) => data[0]);
+            await newReview.$add("tags", tag, { transaction: t });
+            return tag;
```
Tahmeed27, line 64: Here, rather than returning tag, could you return
`{
  name: tag.name
}`
And then this would allow you to directly define tagsRet: Tag[] in line 57 rather than needing to build it in the next few lines

backend/typescript/services/implementations/ReviewService.ts
```diff
@@ -0,0 +1,174 @@
...
+    if (isTest) {
+      this.db = testSequelize;
+    } else {
+      this.db = sequelize;
+    }
+  }
+
+  /* eslint-disable class-methods-use-this */
+  async createReview(review: ReviewRequestDTO): Promise<ReviewResponseDTO> {
+    let publisher: PgPublisher;
```
Tahmeed27, line 39: Why do you keep publisher as a separate variable here? Why not handle it the same way as you did with authors

backend/typescript/services/implementations/ReviewService.ts
```diff
@@ -0,0 +1,174 @@
...
+          byline: newReview.byline,
+          featured: newReview.featured,
+          books: booksRet,
+          tags: tagsRet,
+          updatedAt: newReview.updatedAt.getTime(),
+          publishedAt: newReview.published_at.getTime() / 1000,
+        };
+      });
+    } catch (error: unknown) {
+      Logger.error(`Failed to create entity. Reason = ${error}`);
```
Tahmeed27, line 166: Here, you need to do `${getErrorMessage(error)}`

backend/typescript/services/implementations/ReviewService.ts
```diff
@@ -0,0 +1,174 @@
...
+        return {
+          reviewId: newReview.id,
+          body: newReview.body,
+          coverImages: newReview.cover_images,
+          byline: newReview.byline,
+          featured: newReview.featured,
+          books: booksRet,
+          tags: tagsRet,
+          updatedAt: newReview.updatedAt.getTime(),
+          publishedAt: newReview.published_at.getTime() / 1000,
```
Tahmeed27, line 162: Will this fail if published_at is null?

Tahmeed27 (changes requested):
Amazing work man. This is ham. Made a few minor comments on the code. I'll do a final testing after you update with the PR comments but otherwise looks great!

backend/typescript/services/implementations/ReviewService.ts
obaranek: I needed the publisher variable to add the relation at an outer scope

backend/typescript/services/implementations/ReviewService.ts
obaranek: Looks like the published_at field was required in the model but not in the RequestDTO interface, I'll make the change in the model and update the review service accordingly.

backend/typescript/services/interfaces/IReviewService.ts
```diff
@@ -0,0 +1,70 @@
...
+export interface ReviewResponseDTO {
+  reviewId: number;
+  body: string;
+  byline: string;
+  featured: boolean;
+  // @TODO: uncomment when christine changes are merged
+  // createdBy: number;
+  books: Book[];
+  tags: Tag[];
+  updatedAt: number;
```
Tahmeed27, line 57: Why doesn't the response have a createdAt field?

backend/typescript/services/interfaces/samples/reviewResponse.ts
```diff
@@ -0,0 +1,118 @@
+import { ReviewResponseDTO } from "../IReviewService";
+
+export default [
+  {
+    reviewId: 100,
```
Tahmeed27, line 5: Shouldn't these responses also have an updatedAt and createdAt fields?

backend/typescript/services/interfaces/IReviewService.ts
```diff
@@ -0,0 +1,70 @@
...
+  publishedAt: number | null;
+}
+
+export interface IReviewService {
+  /**
+   * create an Entity with the fields given in the DTO, return created Entity
```
Tahmeed27, line 63:
```suggestion
   * create a review with the fields given in the DTO, return created review
```

backend/typescript/services/interfaces/IReviewService-ti.ts
```diff
@@ -0,0 +1,83 @@
...
+  illustrator: t.opt(t.array("string")),
+  translator: t.opt(t.array("string")),
+  formats: t.array("Format"),
+  minAge: "number",
+  maxAge: "number",
+  authors: t.array("Author"),
+  publishers: t.array("Publisher"),
+  seriesName: t.opt("string"),
+});
+
```
Tahmeed27, line 37: Can this file be moved to a different folder titled interfaceCheckers or something?

backend/typescript/services/interfaces/checkers/IReviewService-ti.ts
```diff
@@ -0,0 +1,83 @@
...
+  seriesName: t.opt("string"),
+});
+
+export const Tag = t.iface([], {
+  name: "string",
+});
+
+export const ReviewRequestDTO = t.iface([], {
+  body: "string",
+  coverImages: t.array("string"),
```
Tahmeed27, line 44: I think you need to re-run the codegen since you changed the models

Tahmeed27 (changes requested):
Okay this is almost perf just found one bug that should be addressed: 
image
It seems to be creating duplicate entries even if they already exist on the table. Could you see how to fix this, also check for this in your test files. And one more thing could you update the codegen and add to the readme saying that you're using ts-interface-checker library, give the command to run code-gen and under what circumstances we would need to run this code-gen (i.e. if we update interface types for eg)

Tahmeed27 (approved):
This is beautiful

PR #27, opened 2021-11-28
Added Tags Component

chriistine:
## Notion ticket link

[Tags component](https://www.notion.so/uwblueprintexecs/Tags-Component-dd408b6f5c074c13965ef5061abbae14)

## Implementation description
* Added endpoint to get all tags
* Added endpoint to delete tag given id
* Added frontend tag component
* Added react-select
* Updated theme
* Added new route - /create-review

**NOTE: searching/creating tag, setting tag on review is listed in the product requirements but was not implemented b/c it needs review service to be done. (perhaps that component of it could be a new ticket?)

## Steps to test
1. Go to /create-review to see tags component in action

* 

## Checklist


obaranek (commented):
I tested the backend. Looks good! I would recommend adding jest tests to make testing easier in case we do some changes in the future.

Tahmeed27 (changes requested):
Okay proposed some minor changes to the code, here and there.

The component looks great btw! But there's a few things that I think might need to be changed:
- Rather than deleting a tag when the cross-button is clicked, the cross-button should only remove it from the tags array in the state. There should be a trashcan icon next to each item in the dropdown list that, when clicked, will delete from db.
- When adding a new tag, change the text to "Add tagName" rather than "Create tagName". And what this should do is add that tag to the tags array (no backend call, just display it as one of the added tags)
- When typing in the form field, I noticed that there's no filtering involved so stuff like typing in "hap" should show the tag named "happy" but not the tag named "fiction". You can search up filtering based on name and you can find a few resources on implementing this
- Is there a way to make it so that when nothing is typed into the text field and the drop down button is clicked, all the tags are shown rather than no tags?


So I also noticed that you didn't use the Notion-style component that was linked in your ticket. It's completely fine if you wanna keep using this component - i just wanted to make sure that you know that you could also look into using that if you wanted: https://codesandbox.io/s/648uv

backend/typescript/middlewares/validators/tagValidators.ts
```diff
@@ -0,0 +1,15 @@
+import { Request, Response, NextFunction } from "express";
+import { getApiValidationError, validatePrimitive } from "./util";
```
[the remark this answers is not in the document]
chriistine: oops i forgot to remove it lol

frontend/src/APIClients/TagAPIClient.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+const deleteTagById = async (id: string): Promise<TagResponse> => {
+  try {
+    const { data } = await baseAPIClient.delete(`/tags/${id.toString()}`, {
+      headers: { Authorization: BEARER_TOKEN },
+    });
+    return data;
+  } catch (error) {
+    return error as TagResponse;
```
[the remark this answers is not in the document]
chriistine: o b/c it was giving me an error otherwise 😅

frontend/src/components/Tags.tsx
```diff
@@ -0,0 +1,77 @@
...
+    },
+  }),
+};
+
+const Tags = (): React.ReactElement => {
+  const [tags, setTags] = useState<TagResponse[]>([]);
+
+  const getTags = () =>
+    new Promise<TagResponse[]>((resolve) => {
+      setTimeout(() => {
```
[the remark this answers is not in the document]
chriistine: https://stackoverflow.com/questions/52984105/react-select-async-loadoptions-is-not-loading-options-properly 
https://react-select.com/async (async multiselect)

Tayef-Shah:
I'm not tryna plug my PR, but like just FYI: 
https://github.com/uwblueprint/ccbc/pull/31 branches off of this one, so I think it makes sense to merge that one into this before merging this to development

chriistine:
> I'm not tryna plug my PR, but like just FYI: #31 branches off of #27 (this branch), so I think it makes sense to merge 31 into here before merging this to development

Yep no I think that makes sense too!

frontend/src/components/common/ConfirmationModal.tsx
```diff
@@ -0,0 +1,72 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
Tahmeed27, line 1:
```suggestion

```

frontend/src/components/common/ConfirmationModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+interface Props {
+  showModal: boolean;
+  onClose: any;
+  handleDelete: any;
+  itemToDelete?: any;
+  deleteType: string;
+}
+
+function ConfirmationModal(props: Props) {
```
Tahmeed27, line 24:
```suggestion
const ConfirmationModal = (props: Props): React.ReactElement => {
```

frontend/src/types/TagTypes.ts
```diff
@@ -0,0 +1,4 @@
+export type TagResponse = {
+  label: string;
```
Tahmeed27, line 2: As per my previous comment, change this to id: string; name: string

frontend/src/components/Tags.tsx
```diff
@@ -0,0 +1,131 @@
...
+    tagAPIClient.getTags().then((newTags) => {
+      setTagOptions(newTags);
+      setIsLoading(false);
+    });
+  }, []);
+
+  const handleCreate = () => {
+    setIsLoading(true);
+
+    // TODO: Create and set tag
```
Tahmeed27, line 57: Did you forget to implement this? The way create should work is it should be strictly frontend only so you can add a new tag to the array: TagOptions

frontend/src/components/pages/CreateReviewPage.tsx
```diff
@@ -0,0 +1,22 @@
+import React from "react";
+
+import AddNewInput from "../AddNewInput";
+import Tags from "../Tags";
+
+const CreateReview = (): React.ReactElement => {
```
Tahmeed27, line 6: Is there a way to do it so that the CreateReview has the state: i.e. array of tags and array of input fields? It's because when we're actually doing the page, we want the actual state to all be stored in this create review page so we can submit all that data at once with a submit button

frontend/src/components/Tags.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const CustomOption = (props: any) => {
+    const { children, data } = props;
+
+    return (
+      <components.Option {...props}>
+        {children}
+        <IconButton
+          aria-label="Delete Button"
+          size="xs"
+          icon={<CloseIcon />}
```
Tahmeed27, line 97: Super small nit but do you think you could change the delete icon to a trashcan icon?

Tahmeed27 (changes requested):
Great work! Deleting and Selecting tags works as expected. Adding new input fields works as expected. Made a few minor comments, and I think you didn't do anything for create tag yet so left a comment about that

frontend/src/components/Tags.tsx
chriistine: Oh I put it as the X icon b/c I asked Prannoy about it and he suggested putting the X rather than the trashcan.

Tahmeed27: Oh ok then it's fine

Tahmeed27 (changes requested):
Nice everything looks great! Really good work. Just noticed one thing that if we try to delete a tag that has been added using the createTag feature, it doesn't exist in the db yet so we shouldn't make a backend call to delete that tag. We should only make backend calls for tags that existed in the db to begin with. When you call get Tags  on initialization, maybe you could store the received tags so that you can check against that in your handle delete function to decide whether you need to make a backend call or not

Tahmeed27 (approved):

PR #35, opened 2021-12-27
Delete Review Endpoint

TheRealJessicaLi:
## Notion ticket link

[Delete Review](https://www.notion.so/uwblueprintexecs/Review-Service-Delete-Review-db722bf7adbf4b18b85cbe55e1b30cdd)

## Implementation description
* Added endpoint to delete review by ID
* Added test
* Added Swagger docs

## Steps to test
1. Verify that all the tests are passing
2. Add some reviews to DB and try deleting them

* 

## Checklist


Tahmeed27 (changes requested):
Hey nice work it seems to be working on all cases except one. For some reason, the publishers table is not being emptied. If I just create one review and then delete that review, all other tables like authors, tags, series etc get cleared as expected but publishers table still has its entry

backend/typescript/services/implementations/__tests__/reviewService.test.ts
```diff
@@ -95,4 +95,59 @@ describe("pg reviewService", () => {
...
+    await expect(reviewService.getReview(deletedId.toString())).rejects.toThrow(
+      `Review id ${deletedId} not found`,
+    );
+
+    // Get updated list of reviews in DB
+    const newResult: ReviewResponseDTO[] = await reviewService.getReviews();
+    expect(newResult.length).toEqual(oldResult.length - 1);
+
+    // Sort actual and expected responses to make sure they are in the same order
+    newResult.sort((a, b) => a.body.localeCompare(b.body));
```
obaranek, line 119: What is this doing? Is it sorting the reviews based on the body?

TheRealJessicaLi: This is sorting the results by the 'body' field so we can ensure the ordering of the sample response matches with the actual response (and we can then check that each element in the actual response is correct). I added a comment in the sample request/response file as well to make sure that the body field is a unique number for testing purposes.

TheRealJessicaLi:
> Hey nice work it seems to be working on all cases except one. For some reason, the publishers table is not being emptied. If I just create one review and then delete that review, all other tables like authors, tags, series etc get cleared as expected but publishers table still has its entry

Just pushed some changes to fix that issue! The problem was the deletion condition for authors/publishers, which was originally a check that there is 1 entry in the intermediate BookAuthor/BookPublisher table for the author/publisher ID. Instead, what we actually want to check is that BookAuthor/BookPublisher entries corresponding to each author/publisher have book IDs that are a subset of the book IDs related to the specific review we are deleting. Consider a case where we have a review with 2 books (which we are deleting asynchronously) with the same authors and publishers. With the original deletion condition, it's possible that when we are checking the number of BookAuthor/BookPublisher entries, the number returned will be greater than 1 for every case (if neither of the books have been deleted at the time of the checks) - this results in the authors/publishers never being deleted.

Tahmeed27 (approved):
Great work! Looks good to me

backend/typescript/services/implementations/__tests__/reviewService.test.ts
obaranek: Ahh that's why my tests were failing. Thank you!

PR #41, opened 2022-01-30
[WIP] Reviewer Association

mahadzaryab1:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d)

## Implementation description
* 

## Steps to test
1.

* 

## Checklist

### DIFF
--- a/backend/typescript/services/implementations/reviewService.ts
+++ b/backend/typescript/services/implementations/reviewService.ts
@@ -19,6 +19,7 @@ import {
   Tag,
   BookResponse,
   AuthorResponse,
+  User,
 } from "../interfaces/IReviewService";
 
 const Logger = logger(__filename);
@@ -196,7 +197,7 @@ class ReviewService implements IReviewService {
       body: review.body,
       byline: review.byline,
       featured: review.featured,
-      createdBy: review.created_by_id,
+      createdByUser: this.getUserDetails(review),
       books,
       tags,
       updatedAt: review.updatedAt.getTime(),
@@ -205,6 +206,16 @@ class ReviewService implements IReviewService {
         : null,
       createdAt: review.createdAt.getTime(),
     };
+  }
+
+  static getUserDetails(review: PgReview): User {
+    const user: User = {
+      id: review.created_by.id,
+      firstName: review.created_by.first_name,
+      lastName: review.created_by.last_name,
+    };
+
+    return user;
   }
 
   async getReview(id: string): Promise<ReviewResponseDTO> {
@@ -221,7 +232,6 @@ class ReviewService implements IReviewService {
         if (!review) {
           throw new Error(`Review id ${id} not found`);
         }
-
         return ReviewService.pgReviewToRet(review);
       });
     } catch (error: unknown) {
--- a/backend/typescript/services/interfaces/IReviewService.ts
+++ b/backend/typescript/services/interfaces/IReviewService.ts
@@ -55,6 +55,12 @@ export interface Tag {
   name: string;
 }
 
+export interface User {
+  id: number;
+  firstName: string;
+  lastName: string;
+}
+
 export interface ReviewRequestDTO {
   body: string;
   byline: string;
@@ -70,7 +76,8 @@ export interface ReviewResponseDTO {
   body: string;
   byline: string;
   featured: boolean;
-  createdBy: number;
+  createdBy?: number | null;
+  createdByUser?: User | null;
   books: BookResponse[];
   tags: Tag[];
   updatedAt: number;


backend/typescript/middlewares/auth.ts
```diff
@@ -40,27 +40,23 @@ export const isAuthorizedByRole = (roles: Set<Role>) => {
 
 /* Determine if request for a user-specific resource is authorized based on accessToken
  * validity and if the userId that the token was issued to matches the requested userId
- * Note: userIdField is the name of the request parameter containing the requested userId */
-export const isAuthorizedByUserId = (userIdField: string) => {
```
susheelk, line 44: Is there a reason this was reformatted? Doesn't look like there are any code changes

backend/typescript/middlewares/auth.ts
```diff
@@ -87,3 +83,37 @@ export const isAuthorizedByEmail = (emailField: string) => {
...
+ *
+ * @param userIdField - the name of the userId field in the request body
+ */
+export const isAuthorizedByUserIdFromBody = (userIdField: string) => {
+  return async (
+    req: Request,
+    res: Response,
+    next: NextFunction,
+  ): Promise<Response | void> => {
+    const userId = String(req.body[userIdField]);
```
nimwijetunga, line 99: Can we just do `const userId = req.params[userIdField] || String(req.body[userIdField]);` and just have one function?

nimwijetunga: Might have to be careful of null access when doing something like `req.params[userIdField]` though

backend/typescript/rest/reviewRoutes.ts
```diff
@@ -9,15 +9,19 @@ import {
 import { getErrorMessage, sendErrorResponse } from "../utilities/errorResponse";
```
nimwijetunga, line 9: One of the ticket requirements is that an admin should not be able to modify the `createdBy` field for another review in a patch. Do we do that in this pr or will that be done in a different one?

nimwijetunga (changes requested):
minor stuff looks good overall

backend/typescript/rest/reviewRoutes.ts
susheelk: Looks like this feature is blocked by https://www.notion.so/uwblueprintexecs/16fb020b93f641c8b384f76a1b3c769e?v=aedc8d99cd8b437cb556ece10f3c0e3c&p=5c6374c816774d9688ea7b4d7aa6e4c0. We could either wait till that gets merged, rebase and then implement this or spin it off into a separate ticket (probably preferable?)

mahadzaryab1: probably better to spin it off into a separate ticket so this PR doesn't go stale

backend/typescript/middlewares/auth.ts
mahadzaryab1: Good idea! I refactored this as you suggested and the functionality remains the same.

backend/typescript/middlewares/auth.ts
```diff
@@ -40,27 +40,23 @@ export const isAuthorizedByRole = (roles: Set<Role>) => {
 
 /* Determine if request for a user-specific resource is authorized based on accessToken
```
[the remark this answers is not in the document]
mahadzaryab1: Could you elaborate a bit more? I'm a bit unsure as to what to update here.

nimwijetunga (approved):
lgtm!

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -207,6 +208,16 @@ class ReviewService implements IReviewService {
     };
   }
 
+  static getUserDetails(review: PgReview): User {
```
dhmi19, line 211: haha same as my PR, last term we didn't write any documentation for the functions. We should probably start this term though to explain what the function does. 

It's a nit pick too but you might wanna add some documentation here to set the standard for future (perhaps more complex) function?

backend/typescript/services/interfaces/IReviewService.ts
```diff
@@ -70,7 +76,8 @@ export interface ReviewResponseDTO {
   body: string;
   byline: string;
   featured: boolean;
-  createdBy: number;
+  createdBy?: number | null;
+  createdByUser?: User | null;
```
dhmi19, line 80: Is there a reason we are returning both the User object and the createdBy number (which I am assuming is the user id?) here? Looking at the notion ticket, I think the sample response only includes the createdByUser object:
`HTTP 200
{
	"id": x,
	...
	"created_by_user": {
		"id": user id
		"name": admin name
	},
	...
}`

I think returning only the userId is the more restful way of doing it, but if we stick with the user object, not sure why we need both fields here

dhmi19 (commented):
Looks good to me overall. Left a few minor comments, mainly asking for clarifications on why certain changes were made so I understand the reasons before giving my approval.

backend/typescript/services/implementations/reviewService.ts
mahadzaryab1: good point - didn't realize i missed this!

backend/typescript/services/interfaces/IReviewService.ts
mahadzaryab1: If you take a look at the logic in the /GET and /POST endpoints, you'll see that `createdBy` is returned in the /POST request and `createdByUser` is returned in the /GET request. Since both request handlers were using the same DTO, I used this workaround to return one or the other.

backend/typescript/middlewares/auth.ts
```diff
@@ -87,3 +83,20 @@ export const isAuthorizedByEmail = (emailField: string) => {
...
   };
 };
+
+/**
+ * isAuthorizedByUserIdFromParameter verifies that the userId passed in the request
+ * body or query parameter matches with the logged in user
+ *
+ * @param userIdField - the name of the userId field in the request body
+ */
+export const isAuthorizedByUserIdFromParameter = (userIdField: string) => {
```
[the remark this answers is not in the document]
mahadzaryab1: Fixed! I initially created two separate functions (one for query parameters and one for request body), but then ended up combining the two.

backend/typescript/services/interfaces/IReviewService.ts
dhmi19: Oh right that makes sense. Minor concern but as far as I know, the more restful way of doing this would be to only return the ID in both cases and if the user needs the user data, they would make another request. I don't think it is worth the changes needed outside the scope of this ticket if we make this change though so I am cool with it

mahadzaryab1: cc @susheelk as this was in the ticket requirements

dhmi19 (approved):
Lgtm!

backend/typescript/services/interfaces/IReviewService.ts
susheelk: @dhmi19 ordinarily I would agree, we should only return the exact data we need and just ids for relationships, but a specific user requirement when displaying a review is to show the name of its creator. (ie we'd never display a review without showing who wrote it). I thought it not worth always sending two requests here

dhmi19: makes sense! I am good with it

susheelk (approved):
Looks good to merge!

PR #42, opened 2022-02-10
Omar/update review service

obaranek:
## Notion ticket link

[Ticket ](https://www.notion.so/uwblueprintexecs/Edit-a-review-4454f563305046e0940842a8b3b08627)

## Implementation description
*  Added the update review service.

## Steps to test
1. Create a review using postman and then update it and verify that the field match as expected.

* 

## Checklist


backend/typescript/rest/reviewRoutes.ts
```diff
@@ -80,18 +80,24 @@ reviewRouter.get(
   },
 );
 
-reviewRouter.delete(
-  "/:id",
-  isAuthorizedByRole(new Set(["Admin"])),
```
mahadzaryab1, line 85: why is this middleware being removed?

backend/typescript/rest/reviewRoutes.ts
```diff
@@ -80,18 +80,24 @@ reviewRouter.get(
...
-  async (req, res) => {
+reviewRouter.put("/:id", reviewRequestDtoValidator, async (req, res) => {
```
mahadzaryab1, line 83: can we add the `isAuthorizedByRole` and `isAuthorizedByUserId` middleware here?

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -32,6 +37,156 @@ class ReviewService implements IReviewService {
     if (db !== sequelize) sequelize.close(); // Using test db instead of main db
   }
 
+  /* eslint-disable class-methods-use-this, no-await-in-loop */
+  async findOrCreateTag(
```
mahadzaryab1, line 41: nit: add function documentation to all functions in this file

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -32,6 +37,156 @@ class ReviewService implements IReviewService {
...
+    await review.$add("tags", newTag, { transaction: t });
+    return newTag;
+  }
+
+  async findOrCreateTags(
+    review: PgReview,
+    tags: TagRequest[],
+    t: Transaction,
+  ): Promise<TagResponse[]> {
+    const tagsRet: TagResponse[] = [];
```
mahadzaryab1, line 59: does it make sense for this to be a `const`? we're adding to the list so the list is changing

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -32,6 +37,156 @@ class ReviewService implements IReviewService {
...
+    for (let i = 0; i < tags.length; i += 1) {
+      const tag = await this.findOrCreateTag(review, tags[i], t);
+      tagsRet.push({ id: tag.id, name: tag.name });
+    }
+    return tagsRet;
+  }
+
+  async findOrCreateSeries(
+    seriesName: string | null | undefined,
```
mahadzaryab1, line 68: why can this be null or undefined?

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -32,6 +37,156 @@ class ReviewService implements IReviewService {
...
+  async findOrCreateTag(
+    review: PgReview,
+    tag: TagRequest,
+    t: Transaction,
+  ): Promise<PgTag> {
+    const newTag = await PgTag.findOrCreate({
+      where: { name: tag.name },
+      transaction: t,
+    }).then((data) => data[0]);
+    await review.$add("tags", newTag, { transaction: t });
```
mahadzaryab1, line 50: what happens if a tag already exists?

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -32,6 +37,156 @@ class ReviewService implements IReviewService {
...
+        title: book.title,
+        series_id: series?.id || null,
+        series_order: book.seriesOrder || null,
+        illustrator: book.illustrator || null,
+        translator: book.translator || null,
+        formats: book.formats,
+        age_range: [book.minAge, book.maxAge],
+      },
+      { transaction: t },
+    );
```
mahadzaryab1, line 138: can we break this block into a helper function?

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -32,6 +37,156 @@ class ReviewService implements IReviewService {
...
+        book.authors[index],
+        t,
+      );
+      authorsRet.push({
+        id: author.id,
+        fullName: author.full_name,
+        displayName: author.display_name || null,
+        attribution: author.attribution || null,
+      });
+    }
```
mahadzaryab1, line 153: same

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -32,6 +37,156 @@ class ReviewService implements IReviewService {
...
+        newBook,
+        book.publishers[i],
+        t,
+      );
+      publishersRet.push({
+        id: publisher.id,
+        fullName: publisher.full_name,
+        publishYear: publisher.publish_year,
+      });
+    }
```
mahadzaryab1, line 167: same

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -32,6 +37,156 @@ class ReviewService implements IReviewService {
...
+      formats: newBook.formats,
+      minAge: newBook.age_range[0].value,
+      maxAge: newBook.age_range[1].value,
+      authors: authorsRet,
+      publishers: publishersRet,
+      series: {
+        id: series?.id || null,
+        name: series?.name || null,
+      },
+    };
```
mahadzaryab1, line 187: same

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -399,6 +480,266 @@ class ReviewService implements IReviewService {
...
+          .filter((oldId: number) => !newTagIds.includes(oldId));
+        await Promise.all(
+          droppedTags.map(async (tagId: number) => {
+            const dbTag = await PgTag.findByPk(tagId, { transaction: t });
+            if (!dbTag) {
+              throw new Error(`Tag id ${dbTag} not found`);
+            }
+            reviewToUpdate.$remove("tags", dbTag, { transaction: t });
+          }),
+        );
```
mahadzaryab1, line 515: can we move each one of these blocks into separate helper functions?

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -399,6 +480,266 @@ class ReviewService implements IReviewService {
...
+        const droppedTags = allReviewTags
+          .map((reviewTag: PgReviewTag) => reviewTag.tag_id)
+          .filter((oldId: number) => !newTagIds.includes(oldId));
+        await Promise.all(
+          droppedTags.map(async (tagId: number) => {
+            const dbTag = await PgTag.findByPk(tagId, { transaction: t });
+            if (!dbTag) {
+              throw new Error(`Tag id ${dbTag} not found`);
+            }
+            reviewToUpdate.$remove("tags", dbTag, { transaction: t });
```
mahadzaryab1, line 513: can we keep the tags and remove the relationship like for the rest of them?

mahadzaryab1 (commented):
added a few comments

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -399,6 +480,266 @@ class ReviewService implements IReviewService {
...
+  async updateReviews(id: string, entity: ReviewRequestDTO): Promise<void> {
+    try {
+      await this.db.transaction(async (t) => {
+        const reviewToUpdate = await PgReview.findByPk(id, {
+          transaction: t,
+          include: [{ all: true, nested: true }],
+        });
+
+        if (!reviewToUpdate) {
+          throw new Error(`Review id ${id} not found`);
```
mahadzaryab1, line 493: are these thrown errors caught somewhere?

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -399,6 +480,266 @@ class ReviewService implements IReviewService {
 
     return result;
   }
+
+  async updateReviews(id: string, entity: ReviewRequestDTO): Promise<void> {
+    try {
```
dhmi19, line 485: Also couldn't update the featured field (I tried changing it from the default (true) to false and it didn't reflect on the db)

dhmi19, line 485: Also couldn't update the created_by field either

dhmi19: Can you also check that the cover image is updated? I tried replacing the default text for the coverImage with another text and the field did not update in the books table row

backend/typescript/rest/reviewRoutes.ts
obaranek: I must've missed it in the merge conflict resolution. I'll them back

backend/typescript/rest/reviewRoutes.ts
obaranek: I must've missed it in the merge conflict resolution. I'll add them back

backend/typescript/services/implementations/reviewService.ts
dhmi19: also the translator field. I couldn't add another translator to the review

backend/typescript/services/implementations/reviewService.ts
obaranek: The const refers to the reference to the array which is a constant. Adding elements to the array doesn't change the reference.

backend/typescript/services/implementations/reviewService.ts
obaranek: To allow the frontend to either not define the field or pass it with a null value.

dhmi19 (changes requested):

dhmi19:
Right now, I can make authors an empty array to remove them from the review. Should we throw an error if a review is edited such that it makes the book have no authors?

obaranek:
> 

I think not having authors for a review is fine, if that's what they want?

backend/typescript/services/implementations/reviewService.ts
obaranek: What would the helper function do? We would still need to call a create book function where we would still pass the fields as above so we won't be really abstracting functionality.

backend/typescript/services/implementations/reviewService.ts
obaranek: what do you mean?

backend/typescript/services/implementations/reviewService.ts
obaranek: Yes, they're caught by the routers.

backend/typescript/services/implementations/reviewService.ts
obaranek: Yea, I was thinking about that before. I think we should just return before. I'll do that for all findOrCreate functions.

backend/typescript/services/implementations/reviewService.ts
obaranek: We could but are we going to benefit much from it? @susheelk  what do you think?

backend/typescript/services/implementations/reviewService.ts
obaranek: Yea we can do that consistency.

dhmi19:
Can we ever have a situation where we add a book to a series? Say we edit a review and change which series the book belongs to. Say we have the following situation:

Book A: 
- Series 1, Order 1

Book B:
- Series 1, Order 2

Book C:
- Series 2, Order 1


Now, let's say that we edit a review to make Book C join series 1. Currently, we don't check if there is a conflict in the ordering of the books when a new book joins the series. Do we need to?

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -399,6 +480,266 @@ class ReviewService implements IReviewService {
...
+                throw new Error(`Tag id ${tag.id} not found`);
+              }
+              await tagToUpdate.update({ name: tag.name });
+            } else {
+              await this.findOrCreateTag(reviewToUpdate, tag, t);
+            }
+          }),
+        );
+
+        // remove books
```
dhmi19, line 534: Why do we delete books from the book table but not authors? Should we keep the books as redundant data like authors for consistency?

dhmi19:
I might be doing something wrong in the testing here, but I tried to remove tags from a review and then re-add those existing tags to the review but the review_tag table didn't update. I attached a screen recording here in case I did something wrong. One thing I noticed, if you try to re-add those tags without the id then it does update the review_tag; however, that implies that those tags are new right? As in, to add a link between reviews and existing tags, we wouldn't pass the id right. 

https://drive.google.com/file/d/1_U_HpXhAjiwgd3HRuDLIWizc2Dtve-qU/view?usp=sharing

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -399,6 +480,266 @@ class ReviewService implements IReviewService {
...
+            const dbTag = await PgTag.findByPk(tagId, { transaction: t });
+            if (!dbTag) {
+              throw new Error(`Tag id ${dbTag} not found`);
+            }
+            reviewToUpdate.$remove("tags", dbTag, { transaction: t });
+          }),
+        );
+
+        // update tags
+        await Promise.all(
```
dhmi19, line 518: might be wrong here, but do we need this Promise.all() if we are using await in all cases inside the async function? 

`
entity.tags.map(async (tag: TagRequest) => {
  ... a bunch of code which is always awaited...
});
`

Will that not just do the same thing as the Promise.all(). If I understand it correctly, Promise.all() just waits for the all the promises inside to resolve but since you are already using await, the map will always contain the awaited result of the promise. If you remove the await, then the Promise.all() might make more sense?

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -399,6 +480,266 @@ class ReviewService implements IReviewService {
...
+              },
+              transaction: t,
+            });
+
+            await PgBook.destroy({ where: { id: bookId }, transaction: t });
+          }),
+        );
+
+        // update books
+        await Promise.all(
```
dhmi19, line 564: same as this promise.all(). It might not be needed?

dhmi19: just check for one and then apply that to all the Promise.all() lol. Idk if promise.all() needs to be used at all if all the promises inside are being awaited

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -399,6 +480,266 @@ class ReviewService implements IReviewService {
 
     return result;
   }
+
+  async updateReviews(id: string, entity: ReviewRequestDTO): Promise<void> {
```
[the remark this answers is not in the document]
obaranek: fixed

backend/typescript/services/implementations/reviewService.ts
obaranek: fixed

obaranek: fixed

backend/typescript/services/implementations/reviewService.ts
obaranek: fixed

backend/typescript/services/implementations/reviewService.ts
obaranek: fixed

obaranek:
> Can we ever have a situation where we add a book to a series? Say we edit a review and change which series the book belongs to. Say we have the following situation:
> 
> Book A:
> 
>     * Series 1, Order 1
> 
> 
> Book B:
> 
>     * Series 1, Order 2
> 
> 
> Book C:
> 
>     * Series 2, Order 1
> 
> 
> Now, let's say that we edit a review to make Book C join series 1. Currently, we don't check if there is a conflict in the ordering of the books when a new book joins the series. Do we need to?

Good point, @susheelk what do you think?

backend/typescript/services/implementations/reviewService.ts
obaranek: I don't see a lot of cases where an admin would want to delete a book and want to add it later. For the sake of simplicity and time, maybe we should leave it that way? @susheelk What do you think?

obaranek:
> 
I changed the deletion of the tag to delete the relation rather than the entity so it should work now

backend/typescript/services/implementations/reviewService.ts
susheelk: We shouldn't be deleting authors, since an author could have written multiple books (ie one referenced by another review)

backend/typescript/services/implementations/reviewService.ts
susheelk: @obaranek This is important, we need to document this file well lol

backend/typescript/services/implementations/reviewService.ts
susheelk: I think ideally we should for the sake of cleanliness, but we need this PR merged ASAP. Only do so if it takes you <1hr and doesn't require re-testing.

susheelk:
> Can we ever have a situation where we add a book to a series? Say we edit a review and change which series the book belongs to. Say we have the following situation:
> 
> Book A:
> 
>     * Series 1, Order 1
> 
> 
> Book B:
> 
>     * Series 1, Order 2
> 
> 
> Book C:
> 
>     * Series 2, Order 1
> 
> 
> Now, let's say that we edit a review to make Book C join series 1. Currently, we don't check if there is a conflict in the ordering of the books when a new book joins the series. Do we need to?

I don't think we should ever allow books to change series. A book is weakly related to a review (ie a book is associated completely with a review and cannot ever exist on its own). Setting Book C to join series 1 is tantamount to deleting Book C from series 2 and adding a new book (call it book D) with the same info as C to series 2.

obaranek:
> > Can we ever have a situation where we add a book to a series? Say we edit a review and change which series the book belongs to. Say we have the following situation:
> > Book A:
> > ```
> > * Series 1, Order 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Book B:
> > ```
> > * Series 1, Order 2
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Book C:
> > ```
> > * Series 2, Order 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Now, let's say that we edit a review to make Book C join series 1. Currently, we don't check if there is a conflict in the ordering of the books when a new book joins the series. Do we need to?
> 
> I don't think we should ever allow books to change series. A book is weakly related to a review (ie a book is associated completely with a review and cannot ever exist on its own). Setting Book C to join series 1 is tantamount to deleting Book C from series 2 and adding a new book (call it book D) with the same info as C to series 2. Does this jive with our current functionality @obaranek ?

I think Dhruv is concerned that we currently don't have a way to validate duplicate series_orders. In other words, we don't have a constraint that series, series_order columns should be unique. @susheelk

susheelk:
> > > Can we ever have a situation where we add a book to a series? Say we edit a review and change which series the book belongs to. Say we have the following situation:
> > > Book A:
> > > ```
> > > * Series 1, Order 1
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Book B:
> > > ```
> > > * Series 1, Order 2
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Book C:
> > > ```
> > > * Series 2, Order 1
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Now, let's say that we edit a review to make Book C join series 1. Currently, we don't check if there is a conflict in the ordering of the books when a new book joins the series. Do we need to?
> > 
> > I don't think we should ever allow books to change series. A book is weakly related to a review (ie a book is associated completely with a review and cannot ever exist on its own). Setting Book C to join series 1 is tantamount to deleting Book C from series 2 and adding a new book (call it book D) with the same info as C to series 2. Does this jive with our current functionality @obaranek ?
> 
> I think Dhruv is concerned that we currently don't have a way to validate duplicate series_orders. In other words, we don't have a constraint that series, series_order columns should be unique. @susheelk 

Oh true mb good catch. I'll make a separate ticket for that

susheelk:
Tested, and generally looks good!
## Basic Cases
* Editing direct fields of reviews (body, byline, featured, etc) :heavy_check_mark: 
* Adding/removing books :heavy_check_mark: 
* Editing book fields :heavy_check_mark: 
* Cannot edit `reviewId` :heavy_check_mark: 
## Potential edge cases
* CAN edit `createdBy` (Should this be intended?)
   * Setting `createdBy` to a non-existent user returns a `500`
* `series` is a required field but can accept an empty array
* `format` is a required field but is nullable

tbh these may not be related to your ticket but some things I discovered about the service, what do you guys think @obaranek @mahadzaryab1 ?

obaranek:
> Tested, and generally looks good! (Sorry it took so long lol)
> ## Basic Cases
> 
>     * Editing direct fields of reviews (body, byline, featured, etc) heavy_check_mark
> 
>     * Adding/removing books heavy_check_mark
> 
>     * Editing book fields heavy_check_mark
> 
>     * Cannot edit `reviewId` heavy_check_mark
> 
> 
> ## Potential edge cases
> 
>     * CAN edit `createdBy` (Should this be intended?)
>       
>       * Setting `createdBy` to a non-existent user returns a `500`
> 
>     * `series` is a required field but can accept an empty array
> 
>     * `format` is a required field but is nullable
> 
> 
> tbh these may not be related to your ticket but some things I discovered about the service, what do you guys think @obaranek @mahadzaryab1 ?

making createdBy to be read-only after it's set should be an easy fix. If that's the behaviour we need then it's an easy fix. Series should be allowed to be nullable I think since books don't need to in series. And I believe format should be required so it shouldn't be allowed to be nullable. Let me know if I should do these change in this ticket or in a seperate one.

susheelk:
> 



> I might be doing something wrong in the testing here, but I tried to remove tags from a review and then re-add those existing tags to the review but the review_tag table didn't update. I attached a screen recording here in case I did something wrong. One thing I noticed, if you try to re-add those tags without the id then it does update the review_tag; however, that implies that those tags are new right? As in, to add a link between reviews and existing tags, we wouldn't pass the id right.
> 
> https://drive.google.com/file/d/1_U_HpXhAjiwgd3HRuDLIWizc2Dtve-qU/view?usp=sharing

Let me try this case. Tbh we should probably not have tag ids? ie just make the tag uniquely identifiable from its name

susheelk:
> > 
> 
> > I might be doing something wrong in the testing here, but I tried to remove tags from a review and then re-add those existing tags to the review but the review_tag table didn't update. I attached a screen recording here in case I did something wrong. One thing I noticed, if you try to re-add those tags without the id then it does update the review_tag; however, that implies that those tags are new right? As in, to add a link between reviews and existing tags, we wouldn't pass the id right.
> > https://drive.google.com/file/d/1_U_HpXhAjiwgd3HRuDLIWizc2Dtve-qU/view?usp=sharing
> 
> Let me try this case. Tbh we should probably not have tag ids? ie just make the tag uniquely identifiable from its name

Yea this exact thing happened to me. Sending a PUT request twice (with the same tags, other info) also seems to remove the tag on the second try.

We can honestly make a separate ticket to refactor tags, @obaranek thoughts?

mahadzaryab1 (approved):
lgtm!

susheelk (approved):

dhmi19 (approved):
tested the changes I requested and looks good to me :)

PR #44, opened 2022-02-14
Get data for admin dashboard

hyunzeekim:
## Notion ticket link

[Admin Dashboard Table](https://www.notion.so/uwblueprintexecs/Admin-Dashboard-Page-8f4bf28e198b4edd802e5b8247075e2a)

## Implementation description
* This ticket will hook up the dashboard table with the actual backend for reviews rather than dummy data. Call create review and get review end-points in backend to do this.

## Steps to test
1. Create/delete reviews and make sure the reviews are reflected on the admin dashboard table

* code correctness

## Checklist


frontend/src/APIClients/ReviewAPIClient.ts
```diff
@@ -1,3 +1,6 @@
+import { BEARER_TOKEN } from "../constants/AuthConstants";
+import baseAPIClient from "./BaseAPIClient";
+
 export type AuthorRequest = {
```
susheelk, line 4: What do you think about moving these types to `frontend/src/types/...`? I know some API clients do this while some keep their types coupled. We should probably decide on something consistent :thinking:

frontend/src/components/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -27,6 +26,18 @@ type ReviewRow = {
 };
 
 const AdminDashboard = (): React.ReactElement => {
+  const [data, setData] = useState<ReviewResponse[]>([]);
+
+  useEffect(() => {
+    getReviews().then((allReviews: ReviewResponse[]) => {
+      const reviews: ReviewResponse[] = [];
```
susheelk, line 33: Is there a reason you couldn't just do `setData(allReviews)`?

susheelk (changes requested):

susheelk:
The deploy job's failing, you may need to run `yarn fix`

frontend/src/components/pages/AdminDashboard/AdminDashboard.tsx
hyunzeekim: Sorry my bad, I was looking off of `Tags.tsx` which needed `forEach` but for `ReviewResponse` it's sufficient to just do `setData(allReviews)`. Will change!

frontend/src/APIClients/ReviewAPIClient.ts
hyunzeekim: Yeah I think it makes more sense to move these to `frontend/src/types/...` similar to auth and tags.

frontend/src/components/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -27,6 +27,14 @@ type ReviewRow = {
 };
 
 const AdminDashboard = (): React.ReactElement => {
+  const [data, setData] = useState<ReviewResponse[]>([]);
+
+  useEffect(() => {
+    getReviews().then((allReviews: ReviewResponse[]) => {
+      setData(allReviews);
```
susheelk, line 34: Uncaught promise?

susheelk, line 34: What happens if allReviews is undefined here?

susheelk (changes requested):

frontend/src/APIClients/ReviewAPIClient.ts
```diff
@@ -1,81 +1,16 @@
...
-  tags: Tag[];
-  updatedAt: number;
-  publishedAt: number | null;
-  createdAt: number;
-};
+import { BEARER_TOKEN } from "../constants/AuthConstants";
+import { ReviewResponse } from "../types/ReviewTypes";
+import baseAPIClient from "./BaseAPIClient";
+
+const getReviews = async (): Promise<ReviewResponse[]> => {
```
mahadzaryab1, line 5: nit: add function documentation

frontend/src/APIClients/ReviewAPIClient.ts
```diff
@@ -1,81 +1,16 @@
...
+  try {
+    const { data } = await baseAPIClient.get("/reviews", {
+      headers: { Authorization: BEARER_TOKEN },
+    });
+    return data;
+  } catch (error) {
+    return error as ReviewResponse[];
```
mahadzaryab1, line 12: can the error object be cast as a `ReviewResponse`?

frontend/src/components/pages/AdminDashboard/mockData.js
```diff
@@ -1,406 +0,0 @@
-const data = [
```
mahadzaryab1, line 1: do we not need this mock data anymore?

frontend/src/types/ReviewTypes.ts
```diff
@@ -0,0 +1,81 @@
+export type AuthorRequest = {
```
mahadzaryab1, line 1: nit: add documentation for all these types

mahadzaryab1 (commented):
added a few comments/questions

frontend/src/APIClients/ReviewAPIClient.ts
```diff
@@ -1,81 +1,16 @@
...
+  }
+};
+
+export default getReviews;
```
dhmi19, line 16: instead of exporting this function, can we export this function as part of an object like this:
`export default {getReviews};`

This is more consistent with `AuthAPIClient.ts` and makes it easier to add more functions in the file. Also, when you import these functions, the precedent is to import the object as a 'reviewApiClient' and then accessing the function as `reviewApiClient.getReviews()`

frontend/src/APIClients/ReviewAPIClient.ts
```diff
@@ -1,81 +1,16 @@
...
+import { ReviewResponse } from "../types/ReviewTypes";
+import baseAPIClient from "./BaseAPIClient";
+
+const getReviews = async (): Promise<ReviewResponse[]> => {
+  try {
+    const { data } = await baseAPIClient.get("/reviews", {
+      headers: { Authorization: BEARER_TOKEN },
+    });
+    return data;
+  } catch (error) {
```
dhmi19, line 11: can you define the type of this error parameter as unknown:
`} catch (error: unknown) { `? This has been a minor issue in the past as typescript complains that it doesn't know the type of the error object

frontend/src/components/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -12,11 +12,11 @@ import MUIDataTable, {
   CustomHeadLabelRenderOptions,
   MUIDataTableColumn,
 } from "mui-datatables";
-import React from "react";
+import React, { useEffect, useState } from "react";
 
-import { ReviewResponse } from "../../../APIClients/ReviewAPIClient";
+import getReviews from "../../../APIClients/ReviewAPIClient";
```
dhmi19, line 17: Can we change this import to:
`import reviewApiClient from "../../APIClients/ReviewAPIClient.ts";` so it's consistent with other files

In Login.tsx we import AuthApiClient.ts as follows:
`import authAPIClient from "../../APIClients/AuthAPIClient";` and then use it like `authApiClient.function()`

frontend/src/components/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -12,11 +12,11 @@ import MUIDataTable, {
...
+import { ReviewResponse } from "../../../types/ReviewTypes";
 import Author from "./Author";
-import data from "./mockData";
 
 type ReviewRow = {
```
dhmi19, line 21: should this type be in the ReviewTypes file? I am fine with keeping it here but I think it might be a bit cleaner keeping it in the types file?

frontend/src/components/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -27,6 +27,14 @@ type ReviewRow = {
 };
 
 const AdminDashboard = (): React.ReactElement => {
+  const [data, setData] = useState<ReviewResponse[]>([]);
+
+  useEffect(() => {
+    getReviews().then((allReviews: ReviewResponse[]) => {
```
dhmi19, line 33: with the change of import, we would have to change this to `reviewApiClient.getReviews()`. The current use of .then() is completely valid. I prefer using await to make the code a bit more readable. If you want to await the response of getReviews, then you can look into this article: https://devtrium.com/posts/async-functions-useeffect

frontend/src/types/ReviewTypes.ts
```diff
@@ -0,0 +1,81 @@
...
+  fullName: string;
+  displayName?: string | null;
+  attribution?: string | null;
+};
+
+export type AuthorResponse = {
```
dhmi19, line 7: Does AuthorResponse not include the id field?

frontend/src/types/ReviewTypes.ts
```diff
@@ -0,0 +1,81 @@
...
+  translator?: string[] | null;
+  formats: Format[] | null;
+  minAge: number;
+  maxAge: number;
+  authors: AuthorRequest[];
+  publishers: Publisher[];
+  seriesName?: string | null;
+};
+
+export type BookResponse = {
```
dhmi19, line 39: Does BookResponse not return the id field?

dhmi19 (changes requested):
Looks good overall! Just left a few minor comments about code formatting. We should be careful about which fields can be made nullable in the backend and how we deal with that in the front end. 

@susheelk can we ever have a review which has no books associated with it? Currently, we can edit a review so that it has no books. Is this something we need to check for in the front end or restrict in the backend?

frontend/src/components/pages/AdminDashboard/mockData.js
hyunzeekim: No the mock data was temporary until our admin dashboard can support data from our endpoints which is being done in this PR

frontend/src/components/pages/AdminDashboard/AdminDashboard.tsx
hyunzeekim: Hmm I think it might make more sense that this type stays in this file instead of ReviewTypes since it's only being used for the mui datatable row.

frontend/src/components/pages/AdminDashboard/AdminDashboard.tsx
hyunzeekim: Just wondering, is there other reasons for switching from then to await or would it just be for better readability?

dhmi19: I only brought it up for consistency with the rest of the project. This should be perfectly valid too! :)

susheelk:
> Looks good overall! Just left a few minor comments about code formatting. We should be careful about which fields can be made nullable in the backend and how we deal with that in the front end.
> 
> @susheelk can we ever have a review which has no books associated with it? Currently, we can edit a review so that it has no books. Is this something we need to check for in the front end or restrict in the backend?

Theoretically, no? We should add a probably check for this both in the frontend and backend. I'll make a separate ticket for this

dhmi19 (approved):
lgtm :)

dhmi19 (approved):

susheelk (approved):
Looks good! Should be good to merge pending @mahadzaryab1's comments

frontend/src/APIClients/ReviewAPIClient.ts
hyunzeekim: hmm it gives an error when trying to cast the error as `ReviewResponse` instead of `ReviewResponse[]`

frontend/src/types/ReviewTypes.ts
susheelk: I feel like that would be excessive documentation? The field names are self-explanatory

PR #48, opened 2022-02-22
Create Component For Previewing Reviews

mahadzaryab1:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d)

## Implementation description
* 

## Steps to test
1.

* 

## Checklist

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -14,6 +14,7 @@ import CreateReview from "./components/pages/CreateReviewPage";
 import Default from "./components/pages/Default";
 import MagazineReview from "./components/pages/MagazineReview";
 import NotFound from "./components/pages/NotFound";
+import PreviewReviewTest from "./components/pages/PreviewReviewTest";
 import Profile from "./components/pages/Profile";
 import Unauthorized from "./components/pages/UnauthorizedPage";
 import { AUTHENTICATED_USER_KEY } from "./constants/AuthConstants";
@@ -98,6 +99,11 @@ const App = (): React.ReactElement => {
                   path={Routes.DEFAULT_PAGE}
                   component={Default}
                   requiredRoles={[UserRole.Admin, UserRole.Subscriber]}
+                />
+                <Route
+                  exact
+                  path={Routes.PREVIEW_REVIEW_TEST}
+                  component={PreviewReviewTest}
                 />
                 <Route exact path="*" component={NotFound} />
               </Switch>
--- a/frontend/src/components/PreviewReviewModal.tsx
+++ b/frontend/src/components/PreviewReviewModal.tsx
@@ -1,0 +1,123 @@
+import {
+  Box,
+  HStack,
+  Image,
+  ModalBody,
+  ModalCloseButton,
+  ModalContent,
+  ModalHeader,
+  ModalOverlay,
+  Stack,
+  Tag,
+  Text,
+} from "@chakra-ui/react";
+import React from "react";
+
+interface PreviewReviewModalProps {
+  title: string;
+  subtitle: string;
+  writtenBy: string;
+  reviewedBy: string;
+  publisher: string;
+  isbn: string;
+  bookType: string;
+  ageDesciption: string;
+  body: string;
+  tags: string[];
+  coverUrl: string;
+}
+
+const PreviewReviewModal = ({
+  title,
+  subtitle,
+  writtenBy,
+  reviewedBy,
+  publisher,
+  isbn,
+  bookType,
+  ageDesciption,
+  body,
+  tags,
+  coverUrl,
+}: PreviewReviewModalProps): React.ReactElement => {
+  return (
+    <>
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader backgroundColor="#FFF8F3">
+          <HStack spacing="225px">
+            <HStack>
+              <Text fontSize={12} fontWeight={600} fontFamily="DM Sans">
+                Previewing:
+              </Text>
+              <Text fontFamily="DM Sans" fontSize={12} fontWeight={400}>
+                Draft
+              </Text>
+            </HStack>
+            <Text fontFamily="DM Sans" fontSize={12} fontWeight={600}>
+              {title}
+            </Text>
+          </HStack>
+        </ModalHeader>
+        <ModalCloseButton />
+        <ModalBody margin={10}>
+          <Stack spacing={10}>
+            <HStack>
+              <Box marginRight={16}>
+                <Image boxSize="250px" objectFit="contain" src={coverUrl} />
+              </Box>
+              <Box fontFamily="DM Sans">
+                <Stack>
+                  <Box paddingBottom={3}>
+                    <Stack spacing={1}>
+                      <Text fontSize={24} fontWeight={800}>
+                        {title}
+                      </Text>
+                      <Text fontSize={10} as="i">
+                        {subtitle}
+                      </Text>
+                    </Stack>
+                  </Box>
+                  <Box fontSize={12} paddingBottom={3}>
+                    <Stack spacing={1}>
+                      <HStack>
+                        <Text fontWeight={600}>Written By:</Text>
+                        <Text fontWeight={400}>{writtenBy}</Text>
+                      </HStack>
+                      <HStack>
+                        <Text fontWeight={600}>Reviewed By:</Text>
+                        <Text fontWeight={400}>{reviewedBy}</Text>
+                      </HStack>
+                    </Stack>
+                  </Box>
+                  <Box fontSize={12} paddingBottom={1}>
+                    <Stack spacing={1}>
+                      <Text>{publisher}</Text>
+                      <Text>{isbn}</Text>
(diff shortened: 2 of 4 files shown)


frontend/src/App.tsx
```diff
@@ -99,6 +100,11 @@ const App = (): React.ReactElement => {
                   component={Default}
                   requiredRoles={[UserRole.Admin, UserRole.Subscriber]}
                 />
+                <Route
+                  exact
+                  path={Routes.PREVIEW_REVIEW_TEST}
+                  component={PreviewReviewTest}
+                />
```
mahadzaryab1, line 107: TODO: Create ticket to remove this route and other references to this page after integration

frontend/src/components/PreviewReviewModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+  Stack,
+  Tag,
+  Text,
+} from "@chakra-ui/react";
+import React from "react";
+
+/**
+ * The model defining the props for the Preview Review Modal component
+ */
+interface PreviewReviewModalProps {
```
mahadzaryab1, line 19: I decided not to use `ReviewResponseDTO` because the component's props should not be coupled with the API response

frontend/src/components/PreviewReviewModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+                Draft
+              </Text>
+            </HStack>
+            <Text fontSize={12} fontWeight={600}>
+              {title}
+            </Text>
+          </HStack>
+        </ModalHeader>
+        <ModalCloseButton />
+        <ModalBody margin={10}>
```
mahadzaryab1: Thoughts on if this bit should not be within the component and defined by the page itself?

frontend/src/components/PreviewReviewModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+                      <Text>{ageDesciption}</Text>
+                    </Stack>
+                  </Box>
+                  <HStack spacing={2}>
+                    {tags.map((tag: string) => (
+                      <Box key={tag}>
+                        <Tag bgColor="#F6E1A8" size="sm" fontWeight={600}>
+                          {tag}
+                        </Tag>
+                      </Box>
```
mahadzaryab1, line 125: TODO: Create ticket to handle overflow if number of tags goes above 5

frontend/src/components/PreviewReviewModal.tsx
uyxela: Seems a bit tedious to have to pass each of these props individually in the parent element, would it make more sense to accept a single review object as a prop

frontend/src/components/PreviewReviewModal.tsx
uyxela: I feel like the <Modal> tag should be in the component, and isOpen and onClose are passed to the component as props

frontend/src/components/pages/PreviewReviewTest.tsx
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * A test page for testing the PreviewReviewModal
+// TODO: Remove once PreviewReviewModal is integrated
+ */
+const PreviewReviewTest = (): React.ReactElement => {
+  const { isOpen, onOpen, onClose } = useDisclosure();
+  return (
+    <>
+      <Button onClick={onOpen}>Open Modal</Button>
+      <Modal isOpen={isOpen} onClose={onClose} size="3xl">
```
uyxela, line 15: move this inside the component?

uyxela (changes requested):
A few things I think need some consideration:
- Consider passing in a single review object instead of each field individually
- Since a review can have multiple books, an array of books should be passed as props and be displayed by the component (not sure if this has been designed)
- When creating a review, the body of the review is saved as a string with HTML, i.e. `"<p>Review body</p>"`
  - Either `dangerouslySetInnerHTML` or having a readOnly React Quill Editor component could work

Small things:
- The top corners of the modal aren't rounded
- Not sure if the modal should be vertically aligned (check with design?)
- The `<Modal>` tag should probably be inside the component, with isOpen and onClose being passed as props

frontend/src/constants/Routes.ts
```diff
@@ -25,3 +25,6 @@ export const CREATE_REVIEW_PAGE = "/create-review";
 export const AUTH_ACTIONS = "/auth/action";
 
 export const DEFAULT_PAGE = "/default"; // FOR TESTING ONLY
+
+// TODO: Remove once PreviewReviewModal is integrated
+export const PREVIEW_REVIEW_TEST = "/preview-review-test";
```
susheelk, line 30: (minor) Could you make this `/test/preview-review`? Worst case for prod we can disable all `/test/**` routes

susheelk (changes requested):

frontend/src/components/PreviewReviewModal.tsx
mahadzaryab1: I decided against accepting a review object as a prop because that tightly couples this component to the API response of the reviews resource. The props of the component should be independent to the response of the API. The other alternative would be to accept `PreviewReviewModalProps` as a prop but it seems like eslint prefers the parameters be restructured.

frontend/src/components/PreviewReviewModal.tsx
mahadzaryab1: done!

frontend/src/components/PreviewReviewModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+            <HStack spacing={175} fontSize={14}>
+              <HStack>
+                <Text fontWeight={600}>Previewing:</Text>
+                <Text fontWeight={400}>Draft</Text>
+              </HStack>
+              <Text fontWeight={600}>{title}</Text>
+            </HStack>
+          </ModalHeader>
+          <ModalCloseButton />
+          <ModalBody margin={10}>
```
susheelk, line 83: The inner modal content should be a separate component. We'll be reusing this component here: https://www.figma.com/file/kEa0TbJ5VKAmI97eGefXs6/Magazine-Review---Subscriber-Experience?node-id=301%3A3660.

susheelk (changes requested):

frontend/src/components/PreviewReviewModal.tsx
susheelk: Ideally, we'd have a generic "Review" type that we could pass rather than a DTO, but not sure what the best practice is

frontend/src/components/PreviewReviewModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * The model defining the props for the Preview Review Modal component
+ */
+interface PreviewReviewModalProps {
+  /** The title of the book that is being reviewed */
+  title: string;
+  /** The subtitle of the book that is being reviewed */
+  subtitle: string;
+  /** The name of the author of the book */
+  writtenBy: string;
```
susheelk, line 26: Mb this was not in the ticket, but the intended behaviour (for now) for multiple books is to just take the 0th one and display its info. Is this doable here or should this be a separate PR?

susheelk (changes requested):

frontend/src/components/PreviewReviewModal.tsx
mahadzaryab1: probably better in a separate PR because we don't have the designs rn

frontend/src/components/PreviewReviewModal.tsx
mahadzaryab1: done

susheelk (approved):

frontend/src/components/PreviewReviewModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+  coverUrl,
+  isOpen,
+  onClose,
+}: PreviewReviewModalProps): React.ReactElement => {
+  return (
+    <>
+      <Modal isOpen={isOpen} onClose={onClose} size="2xl" isCentered>
+        <ModalOverlay />
+        <ModalContent fontFamily="DM Sans">
+          <ModalHeader backgroundColor="#FFF8F3">
```
uyxela, line 71: It looks like the upper corners of the modal are square? `borderRadius` might need to be changed for `ModalHeader`

uyxela (changes requested):
Looks good to me – I think the top left and right corners of the modal still need to be rounded

uyxela:
One thing I just thought of – since we're using `dangerouslySetInnerHTML`, it might be a good idea to do some kind of sanitization or check on the backend

https://github.com/cure53/DOMPurify might work

@susheelk

mahadzaryab1:
> Looks good to me – I think the top left and right corners of the modal still need to be rounded

This was addressed in the Slack thread. Essentially, there is a bug in Chakra UI that causes the rounding of the corners to go away after adding a background colour to the header.

uyxela (approved):
Looks great!

PR #52, opened 2022-03-04
Change Password Flow

uyxela:
## Notion ticket link

[Change Password Flow](https://www.notion.so/uwblueprintexecs/Change-Password-Flow-d27aed7ac94242ffb7c4f364fbebb49c)

Screen Shot 2022-03-06 at 1 21 54 PM
Screen Shot 2022-03-04 at 1 09 46 PM

## Implementation description
* Added a new modal that appears when the Change Password button is pressed in the nav bar
* Component allows the user to change their password by entering their old password and new password

## Steps to test
1. Click on the profile icon and Change Password button 
2. Cases such as entering an incorrect password, different passwords for new and confirm password boxes, etc.

## Checklist


dhmi19:
Could you check if you linted the frontend properly? I get these warnings when I compile the code:
image

dhmi19:
Quick note, the password changed asset is the same (similar?) to the asset I added for the forgot password PR. Should we just use the same asset? If so, we should figure out a way to merge appropriately. Either merge my PR first and then pull in that graphic for yours or vice versa. @susheelk

frontend/src/components/common/ChangePasswordModal.tsx
```diff
@@ -0,0 +1,248 @@
...
+    >
+      <ModalOverlay />
+      <ModalContent padding="10px">
+        <ModalHeader>Change password</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody>
+          <Text fontWeight="bold" mb="8px">
+            Old password
+          </Text>
+          <Input
```
dhmi19, line 204: Can you use the PasswordInputField.tsx component here? It is found in the common components folder. Ik the designs didn't include this component, but I think they should to keep the UX consistent based on other password components in the app too? @susheelk @desginers

dhmi19 (changes requested):

frontend/src/components/common/ChangePasswordModal.tsx
susheelk: +1 on this. That component also gives us nice things like the "eye" to toggle visibility, turning the field red on invalidity, etc.

dhmi19 (changes requested):
Looks good to me. Just a few changes with the copy and it should be good to go!

frontend/src/components/common/ChangePassword/ChangePasswordModal.tsx
```diff
@@ -0,0 +1,264 @@
...
+  /** A boolean representing whether the modal is open or not */
+  isOpen: boolean;
+  /** A function that sets isOpen to false */
+  onClose: () => void;
+}
+
+/**
+ * Type for error-related props to the PasswordInputField component
+ */
+type PasswordFeedback = {
```
liu-jeffrey, line 41: Out of curiosity, is there a reason why we went with `type` rather than `interface`?

frontend/src/components/common/ChangePassword/ChangePasswordModal.tsx
```diff
@@ -0,0 +1,264 @@
...
+  const [error, setError] = React.useState<boolean>(false);
+
+  const resetFeedback = () => {
+    setFeedback("");
+    setOldPasswordFeedback({ isInvalid: false, errorMessage: "" });
+    setNewPasswordFeedback({ isInvalid: false, errorMessage: "" });
+    setConfirmPasswordFeedback({ isInvalid: false, errorMessage: "" });
+  };
+
+  const reset = () => {
```
liu-jeffrey, line 83: Maybe rename `reset` to something more specific, like `resetForm` or something?

frontend/src/components/common/NavBar.tsx
```diff
@@ -16,10 +17,12 @@ import authAPIClient from "../../APIClients/AuthAPIClient";
 import logo from "../../assets/ccbc.png";
 import { UserRole } from "../../constants/Enums";
 import AuthContext from "../../contexts/AuthContext";
+import ChangePasswordModal from "./ChangePassword/ChangePasswordModal";
 
 const NavBar = (): React.ReactElement => {
   const { authenticatedUser, setAuthenticatedUser } = useContext(AuthContext);
   const isAdmin = authenticatedUser?.roleType === UserRole.Admin;
+  const { isOpen, onOpen, onClose } = useDisclosure();
```
liu-jeffrey, line 25: nit: rename these variables to be more specific to the change password modal, e.g. `isChangePasswordModalOpen`, etc.

liu-jeffrey: Not really necessary if we are certain that there won't be other modals on the `NavBar`, but just incase I think it would be helpful.

frontend/src/components/common/ChangePassword/ChangePasswordModal.tsx
```diff
@@ -0,0 +1,264 @@
...
+                  setError(true);
+                }
+              });
+          })
+          .catch((e) => {
+            if ((e as FirebaseError).code === "auth/wrong-password") {
+              setOldPasswordFeedback({
+                isInvalid: true,
+                errorMessage: "The password you entered is incorrect.",
+              });
```
liu-jeffrey, line 144: also a nit, personally i don't even know super well how to do this, but I think syntactically, using Async, Await, is a lot cleaner then using Promises and .then()

liu-jeffrey (commented):
Great work Alex, small minor questions and suggestions but nothing seems broken. I was able to successfully test most cases, except the main case of successfully changing the password, cause I can't seem to remember my actual password oops :P

liu-jeffrey (dismissed):
sorry oops i forgot to click "request changes", but yea 😛

frontend/src/components/common/ChangePassword/ChangePasswordModal.tsx
uyxela: no particular reason - I changed it to interface

frontend/src/components/common/ChangePassword/ChangePasswordModal.tsx
```diff
@@ -0,0 +1,264 @@
...
+      if (authenticatedUser?.email) {
+        // Sign in on the frontend side to validate password
+        signInWithEmailAndPassword(auth, authenticatedUser.email, oldPassword)
+          .then((userCredential) => {
+            // Signed in
+            const { user } = userCredential;
+            updatePassword(user, newPassword)
+              .then(() => {
+                // Update successful
+                setSubmitted(true);
```
dhmi19, line 124: Right now, when a user changes their password, they become unauthorized to make any requests. Instead, we need to log them in again programatically so that they can be authenticated again. We can do this by calling the authAPIClient.login() function which will get a new auth token from the backend and set that to local storage

dhmi19 (approved):
Looks good overall. 

Failing Test Case:
1. User logs in 
2. User changes password
3. User tries to make any request (eg: logout) that needs authentication -> this should fail

dhmi19 (changes requested):
Looks good overall.

Failing Test Case:

User logs in
User changes password
User tries to make any request (eg: logout) that needs authentication -> this should fail

dhmi19 (approved):
lgtm! @susheelk what's the verdict on the PasswordChanged.png graphic? It's the same graphic that I had in one of my PRs. We probable shouldn't duplicate assets if not needed. Do you think we can make a ticket to combine both use cases to only use 1 of the graphics?

susheelk (approved):
LGTM!

PR #91, opened 2022-05-19
Age range field bug fix

Saber74:
## Notion ticket link

[Age range field ](https://www.notion.so/uwblueprintexecs/Age-range-field-e27580186fc74348a7408eb7cf3c4a53)

## Implementation description
* Voted for better UX design. Now upon deleting a number field we can see it as empty instead of NaN

## Steps to test
1. Navigate to the Admin Dashboard 
2. Either click “Add Review” or edit an existing review
3. Either click “Add Book”, or if editing, click an existing book
4. Add an audience age range of any value then delete it (for both min and max), min and max fields should be empty instead of displaying NaN  
5. Fill in all other required fields, put 0 as the min audience age range, some value >0 as the max audience range
6. Try to click save, you should be able to save

* The Modal works as intended, can have a min age of 0, and do not see NaN display upon deletion of age. Can not submit form with an empty age

## Checklist

### DIFF
--- a/frontend/src/components/pages/CreateReview/AddNumberInput.tsx
+++ b/frontend/src/components/pages/CreateReview/AddNumberInput.tsx
@@ -13,7 +13,7 @@ type AddNumberInputProps = {
   mb?: number;
   minNum: number;
   maxNum: number;
-  numberInputFieldValue: number;
+  numberInputFieldValue: number | "";
   setNumberField: (n: number) => void;
 };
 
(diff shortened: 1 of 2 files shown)


susheelk (approved):
* Code looks good!
* Tested (tried adding strings, negative values, blank strings and rejected as required) :heavy_check_mark: 

LGTM, Nice work! :smile:

PR #101, opened 2022-06-05
[Tech] Frontend Global Toasts

Saber74:
## Notion ticket link

[Frontend Toasts](https://www.notion.so/uwblueprintexecs/Frontend-Toasts-d85da818e9a7440db41aff7888cebe7f)

## Implementation description
* created a central component that can be called from anywhere to display a chakra ui toast.

## Steps to test
1. Publish a review (verify a toast appears on the bottom right of the screen displaying a success message)
2. Edit a review (verify a toast appears on the bottom right of the screen displaying a success message)

* Time a toast should remain on screen, previously it was 10 seconds which imo feels extreme, it is currently 5 seconds as of this PR
* Info vs Success for publishing/editing reviews? It was previously info 
image
image
* Here is an example of what happens when we have multiple toasts on screen (
image)
the 2nd toast is error just to be able to differentiate between the 2 (in case they overwrote each other). 
* I have also tested navigating to another page while a toast is visible and the toast remains on it's page until it's specified timeout

## Checklist


frontend/src/components/Toast.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+type ToastStatus = "error" | "warning" | "success" | "info" | undefined;
+
+type ToastProps = {
+  title: string;
+  message: string;
+  status: ToastStatus;
+};
+
+const ToastHook = (): ((
```
susheelk, line 12: I think convention is to call custom hooks useSomething, (ie `useToasts`)

frontend/src/components/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -46,42 +46,39 @@ const AdminDashboard = (): React.ReactElement => {
     onClose: onPreviewModalClose,
   } = useDisclosure();
   const [data, setData] = useState<Review[]>([]);
-  const { notifications } = useContext(NotificationContext);
-  const toast = useToast();
   const [isDeleteModalOpen, setIsDeleteModalOpen] = useState(false);
   const [deleteReviewName, setDeleteReviewName] = useState("");
   const [deleteReviewId, setDeleteReviewId] = useState(-1);
   const [selectedReview, setSelectedReview] = useState<Review>({} as Review);
+  useState<string>("");
```
susheelk, line 52: Do we need these `useState` lines?

frontend/src/components/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -46,42 +46,39 @@ const AdminDashboard = (): React.ReactElement => {
...
-
   const getIndex = (id: number) => {
     return data.findIndex((element) => element.reviewId === id);
   };
 
   const deleteReview = async () => {
-    await reviewAPIClient.deleteReviewById(deleteReviewId.toString());
+    try {
+      await reviewAPIClient.deleteReviewById(deleteReviewId.toString());
+      newToast("Review deleted", "Your review has been published", "info");
```
susheelk, line 73: Looks like this message doesn't match the description (typo :smile: ?)

frontend/src/components/Toast.tsx
```diff
@@ -0,0 +1,42 @@
+import { useToast } from "@chakra-ui/react";
+import { useEffect, useState } from "react";
+
+type ToastStatus = "error" | "warning" | "success" | "info" | undefined;
```
susheelk, line 4: It may be a better idea to use an enum here instead

frontend/src/components/pages/CreateReview/CreateReview.tsx
```diff
@@ -55,15 +53,15 @@ const CreateReview = ({ id }: CreateReviewProps): React.ReactElement => {
     onOpen: onOpenBookModal,
     onClose: onBookModalClose,
   } = useDisclosure();
+
   const [currBook, setCurrBook] = useState<Book | null>(null);
+  const newToast = ToastHook();
 
-  const [showDeleteBookModal, setShowDeleteBookModal] = useState<boolean>(
```
susheelk, line 60: Any clue why these lines were changed? Maybe try running the linter

frontend/src/components/Toast.tsx
```diff
@@ -0,0 +1,42 @@
...
+const ToastHook = (): ((
+  title: string,
+  message: string,
+  status: ToastStatus,
+) => void) => {
+  const [state, setState] = useState<ToastProps | null>();
+  const toast = useToast();
+
+  const addToast = (title: string, message: string, status: ToastStatus) => {
+    const newToast: ToastProps = { title, message, status };
```
susheelk, line 21: Since we're only storing one toast as props, do you know what will happen if multiple toasts are emitted at once? (ie does only the latest one get shown, or do they stack). This might be an edge case worth investigating

frontend/src/components/Toast.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const ToastHook = (): ((
+  title: string,
+  message: string,
+  status: ToastStatus,
+) => void) => {
+  const [state, setState] = useState<ToastProps | null>();
+  const toast = useToast();
+
+  const addToast = (title: string, message: string, status: ToastStatus) => {
```
susheelk, line 20: (nitpick) Maybe the toast status should be the first argument

susheelk (changes requested):
Nice work! (this one was a doozy :smile:). Just added a few comments

susheelk:
> Info vs Success for publishing/editing reviews? It was previously info

I think it should be a success imo, but the designers may be more qualified to answer this

susheelk:
Also linting failed, you may need to run `yarn fix`

frontend/src/components/pages/CreateReview/CreateReview.tsx
Saber74: This one is giving me a headache, my vscode keeps changing it on save due to eslint/prettier and I agree with it in terms of style. Not sure why eslint in the docker does not like it.

frontend/src/components/Toast.tsx
Saber74: Here is an example (
image)
the 2nd toast is error just to be able to differentiate between the 2 (in case they overwrote each other). I have also tested navigating to another page while a toast is visible and the toast remains on it's page until it's specified timeout

frontend/src/components/Toast.tsx
Saber74: The type within chakra UI is explicitly `"error" | "warning" | "success" | "info" | undefined` creating an enum is causing type issues when calling the chakra UI toast

Saber74:
Talked with Amanda for now, editing will remain info, publishing a review will become success. Default time before a toast closes will be 3000.

frontend/src/components/pages/CreateReview/CreateReview.tsx
uyxela: is your vscode using a different prettier config? you might have to add a project specific setting to use the prettier config from this project

uyxela (commented):
lgtm, it would be nice if you could add some documentation to the edited methods in ReviewAPIClient (handleReview, publishCreatedReview, publishEditedReview) as a reminder that they will need to be wrapped in try/catch

also there's a linter error that needs to be fixed

frontend/src/components/Toast.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+type ToastStatus = "info" | "warning" | "success" | "error" | undefined;
+
+type ToastProps = {
+  status: ToastStatus;
+  title: string;
+  message: string;
+};
+
+const useToasts = (): ((
```
susheelk, line 12: Could you add a docstring header here. Also could you add some brief documentation throughout this file on how to use this hook?

susheelk (changes requested):

frontend/src/components/Toast.tsx
susheelk: Awesome, thanks for checking!

frontend/src/components/Toast.tsx
susheelk: Makes sense, thanks!

susheelk (approved):
Nice work! :partying_face:

PR #106, opened 2022-06-08
[Bug] Add new author is empty but field persists

williamtran10:
## Notion ticket link

[Add new author is empty but field persists](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d](https://www.notion.so/uwblueprintexecs/Add-new-author-is-empty-but-field-persists-6e6b02c84ef446a497d0b1624d697c71))

## Implementation description
* Adds new numEmptyFields state. Disables button when numEmptyFields > 0.

## Steps to test
1. Create review
2. Add new book
3. Click add new author (since this is empty it should prevent you from adding 2 new authors)
4. Fill in author name and add a new author
5. Add and delete fields as you want to make sure it still works

## Checklist


jessiiepeng:
The code works as expected! However, idk if this should be addressed in this ticket or not (or just create a new bug ticket) but even if the author field is empty, I can still submit the book even though it should be disabled??
Screen Shot 2022-06-10 at 11 14 56 PM

susheelk:
> The code works as expected! However, idk if this should be addressed in this ticket or not (or just create a new bug ticket) but even if the author field is empty, I can still submit the book even though it should be disabled?? Screen Shot 2022-06-10 at 11 14 56 PM

Damn nice catch! We shouldn't be able to submit the book with 0 authors, but looks like this was an issue before this PR, I'll create a separate ticket

jessiiepeng (approved):
LGTM!

PR #113, opened 2022-06-20
Refactor Edit Review Endpoint

uyxela:
## Notion ticket link

[Refactor Edit Review Endpoint](https://www.notion.so/uwblueprintexecs/Refactor-Edit-Review-Endpoint-de2eefe01ef24799b6aa13491ad1f4ca)

## Implementation description
* Add optional transaction parameter to create and delete review functions
* Change update review function to call create and delete review, passing a transaction
* Also refactored several functions to remove eslint disable no-await-in-loop

## Steps to test
1. Test by updating reviews

## Checklist


backend/typescript/services/implementations/reviewService.ts
```diff
@@ -189,7 +192,119 @@ class ReviewService implements IReviewService {
     };
   }
 
-  async deleteReview(id: string): Promise<void> {
+  // if deleteReview is called without parameter txn, it runs in a try catch block
```
susheelk, line 195: Instead of duplicating this code between the two cases (txn provided or not), could you maybe move this into a helper function? It'll make future changes here much easier

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -189,7 +192,119 @@ class ReviewService implements IReviewService {
...
+                  });
+                }
+              });
+            }
+          });
+        }),
+      );
+
+      // Delete tags (if necessary)
+      const deleteTags = Promise.all(
```
susheelk, line 276: This may not relate to your ticket, but could you remove all references to Tags in the insert and delete methods? We're refactoring Tags rn and these changes might cause conflicts

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -427,36 +542,80 @@ class ReviewService implements IReviewService {
...
+  // otherwise, createReview should be called inside a try catch block with t as parameter
+  async createReview(
+    review: ReviewRequestDTO,
+    id?: string,
+    txn?: Transaction,
+  ): Promise<ReviewResponseDTO> {
     let result: ReviewResponseDTO;
 
+    if (txn) {
+      const newReview = await PgReview.create(
```
susheelk, line 555: Same here, could you move this logic into a helper function?

susheelk (changes requested):
Nice work, thanks for the investigation into transactions! I added a few comments.

chriistine (changes requested):

chriistine (changes requested):
Error 1: 
1. Create a review with a book
2. Edit the review by creating another book with the same values but a different book title.
3. Try to publish
4. It fails to save the newly edited review. You get an error that says, "`Cannot read property 'value' of undefined`"
5. The existing review is deleted. <- Should not be happening.

Error 2:
This doesn't happen often but I think when I try to edit the existing review by adding a new book and playing around with the values in the original book, I get the following error:
`insert or update on table \"books\" violates foreign key constraint \"books_series_id_fkey\"
`

Also when the review fails to save, it deletes the review completely. Don't think this should be happening? If we fail to edit an existing review, the review should stay as same as before. It should not be deleted.

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -189,123 +192,117 @@ class ReviewService implements IReviewService {
...
-                if (
-                  bookAuthorIds.every((bookId) => allBookIds.includes(bookId))
-                ) {
-                  // Delete author
-                  await PgAuthor.destroy({
-                    where: { id: [author.id] },
-                  });
-                }
+    const allBookIds = reviewToDelete.books.map((book: PgBook) => book.id);
+    const deletePublishersAndAuthors = Promise.all(
```
susheelk, line 232: None of these operations seem to be using the provided `txn`. Perhaps this explains some of @chriistine's bugs? :thinking:

susheelk (changes requested):
I also ran into a few issues where improper db operations weren't causing the transactions to rollback. Here's one example:
1. Add a review as normal
2. Edit a book in the review to have the same min and max age, click save/publish
3. This causes an error (rightfully so, this is a completely separate bug that Jessie's fixing)
4. The review gets deleted (this shouldn't happen if the entire edit is wrapped in a transaction)

The min/max bug is unrelated, but reviews shouldn't be deleted upon errors.

susheelk (approved):
LGTM, nice work on this! :smile: 
(Wait for @chriistine's review before merging)

chriistine (approved):
All works now!

PR #119, opened 2022-06-26
Fix misordered book reviews

helenellyx:
## Notion ticket link

[Books in Review are misordered](https://www.notion.so/uwblueprintexecs/Books-in-Review-are-misordered-e6ddddcd2dfd472f999bd1f41ca987f0)

## Implementation description
*  The books were stored in reverse order in the review.books array. Therefore, instead of initalizing the currentBook state at 0 and incrementing it to go to the next book, the state starts at the last element of the array, and it is decremented. 

## Steps to test
Add a Review with multiple Books, set different `seriesOrder` s for each Book. In the Add Review Page, the Books should be sorted by their series order. Publish the review, open the preview component, check that:

1. The books are sorted according to their order in the series
2. the next book is shown when clicking the right arrow (except for the last book)
3. the book before is shown when clicking the left arrow (except for the first book)
4. Try clicking on the different dots in the meatballs menu icon (the dots on the bottom), and check that the page shown corresponds to the right book in the series 

* checking the meatball menu corresponds to the right book

## Checklist

### DIFF
--- a/frontend/src/components/PreviewReview/PreviewReview.tsx
+++ b/frontend/src/components/PreviewReview/PreviewReview.tsx
@@ -26,20 +26,20 @@ interface PreviewReviewProps {
  * what the subscribers of the review will see
  */
 const PreviewReview = ({ review }: PreviewReviewProps): React.ReactElement => {
-  const [currentBook, setCurrentBook] = useState<number>(0);
+  const len = review.books.length - 1;
+  const [currentBook, setCurrentBook] = useState<number>(len);
 
   const handleLeftClick = () => {
-    if (currentBook > 0) {
-      setCurrentBook(currentBook - 1);
-    }
-  };
-
-  const handleRightClick = () => {
-    if (currentBook < review.books.length - 1) {
+    if (currentBook < len) {
       setCurrentBook(currentBook + 1);
     }
   };
 
+  const handleRightClick = () => {
+    if (currentBook > 0) {
+      setCurrentBook(currentBook - 1);
+    }
+  };
   return (
     <>
       <VStack spacing={10}>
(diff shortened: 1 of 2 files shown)


jessiiepeng:
I think you should rebase onto development first / merge. I ran into the reviews.map error here but after I merged development into my local branch it worked.

jessiiepeng:
I don't think the sorting works? I added multiple books in a series and it's appearing in the order I added them as books rather than the series order.

Eg.
I added book1, book3, book2, book5, and then book4.
In the create review page, it shows up in order:

Screen Shot 2022-06-26 at 5 12 26 PM

If you press preview in the create review page, it shows up in the order 5 4 3 2 1. (so reversed)

However, in the preview review model on the admin dashboard page, it shows up in the order: 2, 3, 1, 4, 5. (so not in order)

jessiiepeng:
I think this is because when you get back the books in `PreviewReview.tsx` from the review, it's not in the right order:
Screen Shot 2022-06-26 at 5 19 56 PM
^^ You can see this is out of order.

Look into `CreateReview.tsx`, lines 414-421 to see how they sort the books so that they appear in the right order.
`    {/* Current books display (sorted by seriesOrder) */}
            {books
              .sort((a: Book, b: Book) => {
                const seriesOrderA = a.seriesOrder ? a.seriesOrder : 0;
                const seriesOrderB = b.seriesOrder ? b.seriesOrder : 0;

                return seriesOrderA - seriesOrderB;
              })`

So how you could apply it in this scenario is:
1. sorting the books based on the above code ^ and setting it to a new variable such as `const sortedBooks = reviews.books.sort((a: Book, b: Book) => { ...`
2. revert back the changes that you made so its ascending upwards instead of downwards descending (so we don't need the len variable)
3. Then, everytime in the code where it uses `reviews.books`, use your new variable `sortedBooks` instead so like `review.books[currentBook].title` becomes `sortedBooks[currentBook].title`

I think this should work but you might have to play around :)

jessiiepeng (changes requested):
See comments !

helenellyx:
I am going to make a new branch off development, and close this one, and open new PR.

PR #123, opened 2022-06-27
[Feat] Genres - API Changes

chriistine:
## Notion ticket link

[Book Genres - Read API Changes](https://www.notion.so/uwblueprintexecs/Book-Genres-Read-API-Changes-fe5038e3fc6848aa9c7af86952d7e3c6)

## Implementation description
* Created Service, DTO, Router for genre
* Added genres field to BookResponse - upon GET request to /reviews, /review/:id, genres can be seen in the book's response body
* Added Create/Delete genres endpoint

image

## Steps to test
1. Create a book
3. Can console log to see the genres Horror, Fantasy

* 

## Checklist


susheelk (changes requested):
Hey I'm testing this on postman, and it seems to be failing the following cases :thinking: 
### Case 1:
1. Send a POST to `/reviews` to add a book with a genre (status 200, works fine)
2. Send a GET to `/reviews/:id` where `id` belongs to the review we just created
3. `genres` array is populated (as intended)
4. Send another POST to `/reviews`, creating another book with the same genre as in step 1
5. The first review's book's genres is now an empty array (Send a GET to `/reviews` to verify)

## Case 2
1. Add a review containing a book with a genre
2. Edit the review (sent a PATCH to `/reviews:id` to remove the genre
3. Add the same genre again
4. Genres array is now empty :cry:
 
Could you look into these? Thanks!

backend/typescript/services/interfaces/IReviewService.ts
```diff
@@ -24,6 +24,10 @@ export interface PublisherResponse {
   publishYear: number;
 }
 
+export interface Genre {
```
susheelk, line 27: Do we need an explicit `Genre` interface? Since we're only displaying Genre names, we could just use `string[]` everywhere instead of `Genre[]`. 
The reviews API response would look something like:
```json
{
  "id": 1,
  "books": [
    "id": 1,
    ...
    "genres": ["a", "b", "c"], // strings here
  ],
  ...
}

File: backend/typescript/swagger.yml [hunk 2]
```diff
@@ -1054,4 +1054,29 @@ paths:
...
+              schema:
+                type: array
+                items:
+                  type: object
+                  properties:
+                    name:
+                      type: string
+        '401':
+          description: Unauthorized Request.
+        '404':
```
susheelk, line 1079: (nit) Doesn't look like we actually encounter a `404` case on this route

susheelk (changes requested):
Code looks good! (just added a few minor comments) Tested, and functionality looks good! :slightly_smiling_face:

backend/typescript/services/interfaces/IReviewService.ts
susheelk: EDIT: probably too much work to

susheelk:
#135 was merged, could you pull development when you have a chance?

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -265,6 +301,20 @@ class ReviewService implements IReviewService {
           }
         });
 
+        // Delete genres (if necessary)
+        book.genres.forEach(async (genre: PgGenre) => {
```
susheelk, line 305: Just tested this. Looks like this block isn't actually deleting the genre when no books are associated with it. We ran into a similar issue with Tags (See #136), so you can just copy that fix (forewarned, it's kinda ugly :cry:).

I think this has to do with async weirdness but not sure :shrug:

susheelk (changes requested):

susheelk (approved):
LGTM! :partying_face:

PR #126, opened 2022-07-04
[Feat] Create user through givecloud subscription

Saber74:
## Notion ticket link

[Webhook Consumer: Create User](https://www.notion.so/uwblueprintexecs/Webhook-Consumer-Create-User-9b8b3f17849a4c40a092e83092bc9db4)

## Implementation description
* Created a new router and an endpoint that can be used to create a new account for any user who purchases a membership through givecloud. 

## Steps to test
1. 

##TODO : 
1) Integration test. 
2) Figure out what to do for the other kinds of subscriptions, as far as am concerned student/senior and individual should not be any different from each other and will just be "Subscriber" the issue comes from the associate/corporate  types 
image

## Checklist


backend/typescript/rest/givecloud.ts
```diff
@@ -0,0 +1,43 @@
+import { Router } from "express";
+import AuthService from "../services/implementations/authService";
+import UserService from "../services/implementations/userService";
+import { sendErrorResponse } from "../utilities/errorResponse";
+import { Role } from "../types";
+
+const givecloudRouter: Router = Router();
+
+givecloudRouter.post("/", async (req, res) => {
+  const { membership } = req.body;
```
susheelk, line 10: (nit) Might be cleaner to do `const  { membership, supporter } = req.body

backend/typescript/rest/givecloud.ts
```diff
@@ -0,0 +1,43 @@
...
+      date.setDate(date.getDate() + membership.days_to_expire);
+      const user = await userService.createUser({
+        firstName: supporter.first_name,
+        lastName: supporter.last_name,
+        email: supporter.email,
+        roleType,
+        subscriptionExpiresOn: date,
+        password: "",
+      });
+      await authService.sendEmailVerificationLink(email);
```
susheelk, line 33: We don't actually use `sendEmailVerificationLink`. This is a holdover from starter code. We've implemented a custom `sendPasswordSetupLink` instead for users to setup their account. 

See the `/register` route for more details. We basically want to replicate that signup flow here, but with the non-Admin roletypes. You may find it useful to move that code into a separate function (Perhaps in the user service? idk) and just call it here.

backend/typescript/rest/givecloud.ts
```diff
@@ -0,0 +1,43 @@
+import { Router } from "express";
+import AuthService from "../services/implementations/authService";
+import UserService from "../services/implementations/userService";
+import { sendErrorResponse } from "../utilities/errorResponse";
+import { Role } from "../types";
+
+const givecloudRouter: Router = Router();
+
+givecloudRouter.post("/", async (req, res) => {
```
susheelk, line 9: (nit) I think convention is to name webhooks `<noun>.<verb>`. Since this hook listens to when user subscriptions are paid, maybe name the route something like `/givecloud/user.subscription_paid`

susheelk (changes requested):
Thanks for investigating this! I added a few comments

backend/typescript/rest/givecloud.ts
```diff
@@ -0,0 +1,56 @@
+import { Router } from "express";
+import AuthService from "../services/implementations/authService";
+import UserService from "../services/implementations/userService";
+import { sendErrorResponse } from "../utilities/errorResponse";
+import { AuthDTO, Role, UserDTO } from "../types";
+
+const givecloudRouter: Router = Router();
+
+const authDtoToToUserDto = (authDTO: AuthDTO): UserDTO => {
```
Saber74, line 9: might be overkill, not sure what we want to return, but ik it's not good to return authDTO given it has some tokens.

susheelk: Yup, makes sense. Technically we only have to return a `201`, but it might be useful for debug purposes to return the `UserDTO` like so

backend/typescript/rest/givecloud.ts
```diff
@@ -0,0 +1,56 @@
...
+  }) => ({ id, firstName, lastName, email, roleType, subscriptionExpiresOn }))(
+    authDTO,
+  );
+  return userDTO;
+};
+
+const userService = new UserService();
+const authService = new AuthService(userService);
+
+givecloudRouter.post("/user.subscription_paid", async (req, res) => {
```
susheelk, line 26: Could you close this route to the public (ie add the `isAuthenticatedByRole("Admin")` middleware) temporarily while we implement the secret verification? Leaving this open may be a security vulnerability

backend/typescript/services/interfaces/authService.ts
```diff
@@ -128,6 +128,24 @@ interface IAuthService {
...
+   * Used to register a new account, it creates a user in the db and then sends an email with set up password to register
+   * @param firstName: first name of the user  to be registered
+   * @param lastName: last name of the user to be registered
+   * @param email the user's email associated with their account
+   * @param roleType: role of the user, one of the predefined roles in the Role type
+   * @param subscriptionExpiresOn the date of the subscription will expire on, null if admin
+   * @returns AuthDTO (token and UserDTO)
+   * @throws Error if unable to register the user and send the registartion email
+   */
+  sendRegistrationEmail(
```
susheelk, line 142: (nit) Could you make this method name more descriptive? (ie `createUserAndSendRegistrationEmail`) It might be useful for callers to know this also creates an account

susheelk (changes requested):
Code looks good! (just minor nitpicks). Will give this a test!

susheelk:
@Saber74 could you pull `development` into this branch?

Saber74:
> @Saber74 could you pull `development` into this branch?

@susheelk  done!

backend/typescript/rest/givecloud.ts
```diff
@@ -0,0 +1,58 @@
...
+  );
+  return userDTO;
+};
+
+const userService = new UserService();
+const authService = new AuthService(userService);
+givecloudRouter.use(isAuthorizedByRole(new Set(["Admin"])));
+
+givecloudRouter.post("/user.subscription_paid", async (req, res) => {
+  try {
```
susheelk, line 29: Could you add some validation here? ie either use the built in validator middleware, or write your own. Maybe return a `400` if any required fields are missing

backend/typescript/rest/givecloud.ts
```diff
@@ -0,0 +1,58 @@
...
+    const { membership, supporter } = req.body;
```
susheelk, line 30: From the example response, isn't "membership" nested within the "supporter" object? This might be why I'm getting the "cannot read property" error

susheelk (changes requested):
(Sorry this took so long :cry:) I tried testing this by sending a request to the endpoint, and I'm getting this error: 
image
I'm guessing it has something to do with accessing `membership.name`

susheelk (changes requested):
Code and functionality looks good! :slightly_smiling_face: 

Found an issue with the validator tho: the route code still runs even if the validator fails:
image
image

I put that "running this" console log on the first line and gave it a request body with no email field. The response is as desired though:
image

Saber74:
> Code and functionality looks good! 🙂
> 
> Found an issue with the validator tho: the route code still runs even if the validator fails: image image
> 
> I put that "running this" console log on the first line and gave it a request body with no email field. The response is as desired though: image

@susheelk oh man that's probably the return that I removed since now res.status would be set twice, what we can do is only proceed if there is no error and add an else with the 400 error. Not a style of code am excited about though so if you have any better ideas shoot them my way

susheelk (approved):
LGTM! Nice work on this, I know it was a pain :cry:

PR #136, opened 2022-07-14
[Feat] Book Tags - API Changes

williamtran10:
## Notion ticket link

[Book Tags - API Changes](https://www.notion.so/uwblueprintexecs/16fb020b93f641c8b384f76a1b3c769e?v=aedc8d99cd8b437cb556ece10f3c0e3c&p=66662814c29843fc9d7bd152f9a40035)

## Implementation description
updated Service, DTO, Router for tags
updated reviewService
updated Create/Delete tag endpoint

## Checklist


williamtran10:
Base branch is currently `william/remove-async-from-delete-review` so that it's easier to see changes relevant to this ticket.

backend/typescript/rest/tagRoutes.ts
```diff
@@ -27,12 +27,12 @@ tagRouter.get(
 );
 
 tagRouter.delete(
-  "/:id",
+  "/:name",
```
susheelk, line 30: I don't think we actually need a delete route on Tags since they're deleted automatically with reviews :thinking:

susheelk (changes requested):
Code looks good! Will give this a test once #135 gets merged

backend/typescript/rest/tagRoutes.ts
williamtran10: Should we keep it around just in case something wonky happens and we need to manually delete a tag?

susheelk: I think this route (and the `deleteTag` function in the TagService) would only work if the Tag to be deleted is not associated with any books (in which case the Review Service would have already implicitly deleted it). I guess there's no harm in leaving it tho. If you're gonna do so, maybe add a comment noting it's only for dev purposes

williamtran10: On second thought, it doesn't make much sense to have it, so it is now removed.

susheelk:
Merge conflicts 😢

susheelk (changes requested):
Hey I gave this a test, I'm able to add Tags just fine, but I'm not able to delete Reviews now. Getting this constraint violations:
image

williamtran10:
LGTM, just fix lint

susheelk (approved):
:partying_face: :balloon: :pray:

PR #144, opened 2022-07-20
[Feat] Add Pagination for Reviews Endpoint

chriistine:
## Notion ticket link

[Reviews Endpoint Pagination](https://www.notion.so/uwblueprintexecs/Reviews-Endpoint-Pagination-f230171da9604ee8be1f7e20dd239a94)

## Implementation description
* Added pagination for GET reviews

* Ensure the frontend still returns the reviews without any errors
* Send GET reviews with both query params: page and size (Ex: /reviews?page=2&size=5) and check the number of reviews we get is correct
* Send GET reviews with just page as query params and check the number of reviews we get is correct
* Send GET reviews with just size as query params and check the number of reviews we get is correct
* Send GET reviews without query params and check the number of reviews we get is correct

## Checklist


backend/typescript/services/interfaces/IReviewService.ts
```diff
@@ -97,6 +97,13 @@ export interface ReviewRequestDTO {
   books: BookRequest[];
 }
 
+export interface RetrieveReviewResponseDTO {
```
susheelk, line 100: (nit) Maybe change this name to `PaginatedReviewResponseDTO`

frontend/src/components/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -56,8 +60,13 @@ const AdminDashboard = (): React.ReactElement => {
 
   useEffect(() => {
     setIsLoading(true);
-    reviewAPIClient.getReviews().then((allReviews: ReviewResponse[]) => {
-      setData(mapReviewResponseToReview(allReviews));
+    reviewAPIClient.getReviews().then((resp: RetrieveReviewResponse) => {
```
susheelk, line 63: This seems to only fetch the first 10 reviews. I think we need to either implement pagination on the admin table (preferred) or set the default behaviour in the backend to fetch all reviews by default (ie default page size is num of total reviews)

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -498,18 +499,38 @@ class ReviewService implements IReviewService {
...
+          limit,
+          offset,
         });
 
-        return reviews.map((r) => ReviewService.pgReviewToRet(r));
+        const currentPage = page ? parseInt(page, 10) : 0;
+        const totalPages = Math.ceil(count / limit);
+
+        return {
+          totalReviews: count,
```
susheelk, line 529: This count seems to be wrong for some reason, and it's throwing off the `totalPages` as well. For example, I only have 5 reviews here in my database but it says I have 11:
image

susheelk (changes requested):
Nice work, and thanks for the quick turnaround! :slightly_smiling_face:. Pagination seems to work, but I found an issue with the total count:

backend/typescript/services/implementations/reviewService.ts
chriistine: Wait that's weird the count seems to be working for me. @Saber74 Can you also try and confirm if the count is off

susheelk: @chriistine I think this might be counting books instead of reviews for some reason, but I'm not too sure :thinking:. Try it with the seed data to recreate. There should be only 7 reviews, but `totalReviews` is 14 for me.

backend/typescript/services/implementations/reviewService.ts
```diff
@@ -498,18 +499,41 @@ class ReviewService implements IReviewService {
...
     try {
       result = await this.db.transaction(async (t) => {
-        reviews = await PgReview.findAll({
+        const limit = size ? parseInt(size, 10) : undefined;
+        const offset = this.getPaginationOffset(page, limit);
+
+        const { rows, count } = await PgReview.findAndCountAll({
           transaction: t,
           include: [{ all: true, nested: true }],
+          limit,
```
susheelk, line 521: @chriistine I tried adding a `distinct: true` field here, and it seemed to solve the incorrect count issue. This may have other side effects tho, as it could be filtering distinct associations (ie tags/genres might show up only for 1 book even if they've been assigned to multiple books), so I'd suggest thoroughly testing if we're going that route :thinking:

susheelk: For reference: https://github.com/sequelize/sequelize/issues/10557

susheelk (approved):
LGTM, nice work! :slightly_smiling_face:

PR #152, opened 2022-08-01
[Feat] Create Filter Box Component

helenellyx:
## Notion ticket link

[Filter Function - Filter Boxes Component](https://www.notion.so/uwblueprintexecs/Filter-Function-Filter-Boxes-Component-6df327e0766947de894a776cdc409665)

## Implementation description
* Note that the design of the filter box is not included in this ticket
* Created the filter box component with two filters applied
* audience: single select filter
* genres: multi-select filter

## Steps to test
1. Instead of passing options to choose from in the dropdown menu of both filters as a prop, hardcode two arrays of options and test with your arrays 
2. verify that the audience is single-select
3. verify that genres is multi-select
4. verify that `genres`  and `age` has the correct value 
5. verify that when clear is clicked, both `genres` and `age` are cleared and the filter box is cleared
6. **note**, the only way to test the apply button is to check that `genres`  and `age` has the correct value, since right now the genres and age filters that the user chose is passed to a function outside of the component which has not been defined 

* the test route is `/test/filter`
* please check that I did not break any functionalities for the genre and tags selection for adding a book

## Checklist

### DIFF
--- a/frontend/src/components/pages/FilterBox.tsx
+++ b/frontend/src/components/pages/FilterBox.tsx
@@ -32,30 +32,49 @@ const FilterBox = ({
     setAgeFilter([]);
   };
   return (
-    <SimpleGrid minChildWidth="140px" spacingX="10px">
-      <AddMultiSelect
-        id="Audience"
-        label=""
-        placeholder="Audience"
-        options={ageOptions}
-        setOptions={null}
-        optionsSelected={age}
-        setOptionsSelected={setAge}
-        allowMultiSelectOption={false}
-        allowAddOption={false}
-      />
-      <AddMultiSelect
-        id="genre"
-        label=""
-        placeholder="Genres"
-        options={genreOptions}
-        setOptions={null}
-        optionsSelected={genres}
-        setOptionsSelected={setGenres}
-        allowMultiSelectOption
-        allowAddOption={false}
-      />
-      <Stack direction="row" align="center" width="100%" h="120%" spacing={1}>
+    <SimpleGrid columns={[1, 4, 4]} spacingX="10px">
+      <Stack
+        direction="row"
+        align="center"
+        w="100%"
+        h="100%"
+        spacing={3}
+        padding={0}
+      >
+        <AddMultiSelect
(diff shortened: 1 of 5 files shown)


frontend/src/components/pages/FilterBox.tsx
```diff
@@ -0,0 +1,68 @@
...
+            label=""
+            placeholder="Genres"
+            options={genreOptions}
+            setOptions={null}
+            optionsSelected={genres}
+            setOptionsSelected={setGenres}
+            allowMultiSelectOption
+            allowAddOption={false}
+          />
+          <Button onClick={clickApply}>Apply</Button>
```
jessiiepeng, line 61: I think these two "Apply" and "Clear" can be centered vertically of the boxes: they're a little more towards the top right now.
Screen Shot 2022-08-03 at 7 17 03 PM

jessiiepeng:
Do we need to make this component responsive in this ticket? In mobile view it looks a bit squished the words
Screen Shot 2022-08-03 at 7 22 48 PM

frontend/src/components/pages/FilterBox.tsx
jessiiepeng: Do we need to make this responsive for mobile view? Looks a bit squished right now
Screen Shot 2022-08-03 at 7 23 32 PM
.

jessiiepeng:
The functionality looks good, I don't think it broke anything for adding tags/genres on books. Just some small styling fixes.

frontend/src/components/pages/FilterBox.tsx
helenellyx: I think @susheelk mentioned that the design of the filter boxes and buttons is not included in this ticket 🤔 I could work on it tho if needed

susheelk: I think it's a good idea to make this somewhat responsive, even if it doesn't have to look good. Maybe have the buttons flow to a new line?

jessiiepeng (approved):
LGTM! yay :)

PR #153, opened 2022-08-01
[Feat] Made givecloud private by an env variable

Saber74:
## Implementation description
* Added a variable to the env file (not pushed)
* Used that variable and double checked with the authorization sent from a post request to make sure they are the same other wise we return a 400 error, and do not proceed further

1. Send a request to the givecloud endpoint without an authorization (should return a 400 erro)
2. Send a request to the givecloud endpoint with the correct authorization key but wrong value (400 error)
3. Send a request to the givecloud endpoint with the correct key and value (goes through)

## Checklist

### DIFF
--- a/backend/typescript/middlewares/givecloud.ts
+++ b/backend/typescript/middlewares/givecloud.ts
@@ -17,4 +17,4 @@ const isGiveCloudEnabled = () => {
   };
 };
 
-export default isGiveCloudEnabled
+export default isGiveCloudEnabled;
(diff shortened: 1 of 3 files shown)


backend/typescript/middlewares/givecloud.ts
```diff
@@ -0,0 +1,20 @@
+import { Request, Response, NextFunction } from "express";
+
+/* 
+checks the .env file to see if the givecloud webhook is enabled,
```
susheelk, line 4: (nit) This checks everywhere environment variables are set, not just the `.env` file. (ie we could set the variable on Heroku too when deployed)

susheelk (approved):
LGTM! (just a minor nitpick)

PR #158, opened 2022-08-08
[Feat] Filter Functionality

chriistine:
## Notion ticket link

[Filter Functionality](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d)

## Implementation description
* Added the following filters: genres, minAge, maxAge, featured, draft
* Ensured search & pagination works with and without filtering

## Steps to test
1. Test with multiple variations of filtering/search/pagination
* Query Example: /reviews?genres=Dystopian,Action&minAge=15&maxAge=20&featured=false&draft=true&size=1

Some specifications:
* As long as one or more of the list of genres in the query are in the review, that review would be selected
* Featured = true only grabs reviews with featured = true, and draft = true only grabs unpublished reviews
* minAge, maxAge: Can specify both, or one, or neither. It grabs all reviews that have overlap of the age ranges specified.

* 

## Checklist


backend/typescript/services/implementations/reviewService.ts
```diff
@@ -550,18 +601,75 @@ class ReviewService implements IReviewService {
...
+        let reviewIds: number[] = [];
+        if (searchedIds.length > 0 && filteredIds.length > 0) {
+          reviewIds = filteredIds.filter((id) => searchedIds.includes(id));
+        } else if (searchedIds.length > 0) {
+          reviewIds = searchedIds;
+        } else if (filteredIds.length > 0) {
+          reviewIds = filteredIds;
+        }
+
+        const searchAndFilterOpt =
```
susheelk, line 659: I think we need to handle the case where both `reviewIds` and `searchIds` return nothing. This could happen if we provide a genre that doesn't exist as a filter. Currently, this bypasses the filter altogether and returns all reviews:
image
(notice that the returned review doesn't have "random" as its genre).

Maybe one approach to doing this would be to distinguish between null/empty list for `searchIds`, `filterIds`. (ie `null` means param wasn't provided. Empty list means that param was provided and filter/search returned nothing)

susheelk (changes requested):
Really awesome work on this! (despite the tedium :smiling_face_with_tear:  :tada:)
Code looks good. Tested and the filters work, just found one bug:

susheelk (approved):
LGTM, Nice work on this! :slightly_smiling_face:

PR #192, opened 2022-11-18
Get Creator Endpoint

aaggarwal10:
## Notion ticket link

[GET Creators Route](https://www.notion.so/uwblueprintexecs/GET-Creators-Route-b0426592dbb54a9a997f166bbf4238c0)

## Implementation description
* Refer to notion ticket for implementation steps.

## Steps to test
1. Make calls via postman:
image

* Type of age range gave me a lot of problems. Initially, I had it typed as Range, but even though I was able to get it into the database, I was no longer able to read it from the database as it would always return undefined. I changed it to a string and did some parsing to make it work. Not sure if this is ideal, but am unsure of how to move forward with the sequelize ORM data type since even just getting it from the table returns undefined.

## Checklist


backend/typescript/rest/creatorRoutes.ts
```diff
@@ -1,11 +1,114 @@
...
 creatorRouter.get("/");
 
-creatorRouter.get("/", creatorDtoValidator, async (req, res) => {
-  res.status(200).send("hello creator");
-});
+const creatorService: ICreatorService = new CreatorService();
+
+function isOverlap(ageRange: string, ageRangeCmp: string): boolean {
+  const lowerInclusiveBound = parseInt(
+    ageRange.split(",")[0].slice(1) || "",
```
Shrinjay, line 14: Not quite sure what this defaults to or what the `||` is for? For one, if anything goes wrong here i.e `ageRange` is not a string or is malformed, then this will throw an error, so the OR condition will never be used. I think it would be better to just get rid of the OR statement entirely and ensure the string is always well formed

backend/typescript/rest/creatorRoutes.ts
```diff
@@ -1,11 +1,114 @@
...
+  "/",
+  isAuthorizedByRole(new Set(["Admin", "Subscriber", "Author"])),
+  async (req, res) => {
+    const { id, location, ageRange, genre, status } = req.query;
+    const isAdmin = !isAuthorizedByRole(new Set(["Admin"]));
+    if (id) {
+      if (typeof id === "string") {
+        const idNumeric = parseInt(id, 10);
+        // Get User By Id
+        if (Number.isNaN(idNumeric)) {
```
Shrinjay, line 47: The validator should take care of this, no need for it here

backend/typescript/rest/creatorRoutes.ts
```diff
@@ -1,11 +1,114 @@
...
+          res
+            .status(400)
+            .json({ error: "id query parameter must be a number." });
+        } else {
+          try {
+            const creator = await creatorService.getCreatorById(id);
+            if (isAdmin && creator.isApproved === false) {
```
Shrinjay, line 54: This filtering logic, checking if the creator is approved, should happen in the service imo

backend/typescript/rest/creatorRoutes.ts
```diff
@@ -1,11 +1,114 @@
...
+          }
+        }
+      } else {
+        res.status(400).json({ error: "id query parameter must be a number." });
+      }
+    } else {
+      // Get all creators with filter
+      try {
+        if (location && typeof location !== "string") {
+          res
```
Shrinjay, line 70: Again, all this validation should happen in a validator, looks messy in here

backend/typescript/rest/creatorRoutes.ts
```diff
@@ -1,11 +1,114 @@
...
+            .status(400)
+            .json({ error: "status query parameter must be a boolean." });
+          return;
+        }
+        const creators = await creatorService.getCreators();
+
+        const filteredCreators = creators.filter(
+          (creator) =>
+            (creator.isApproved || isAdmin) &&
+            (!status || creator.isApproved === (status === "true")) &&
```
Shrinjay, line 100: I'm guessing this is to only use the `status` filtering if status is defined? Why not use a ternary operator here?

Shrinjay (changes requested):
Couple things I noticed with the code structure and how the range comparison was implemented

backend/typescript/rest/creatorRoutes.ts
aaggarwal10: I get this weird parameter error on the id if i remove the if statement tho:

backend/typescript/rest/creatorRoutes.ts
aaggarwal10: Again, because of the type error above.

backend/typescript/rest/creatorRoutes.ts
Shrinjay: `(status ? creator.isApproved === (status === "true") : true)`

backend/typescript/rest/creatorRoutes.ts
```diff
@@ -1,11 +1,82 @@
...
+        try {
+          res.status(200).json(await creatorService.getCreatorById(id));
+        } catch (error: unknown) {
+          sendErrorResponse(error, res);
+        }
+      }
+    } else {
+      // Get all creators with filter
+      try {
+        const creators = await creatorService.getCreators();
```
Shrinjay, line 61: Sorry I didn't catch this earlier, but we 100% need to do all this filtering in the service layer itself, why are we doing it in the routes?

Shrinjay (changes requested):

backend/typescript/rest/creatorRoutes.ts
```diff
@@ -1,11 +1,79 @@
...
+        const filteredCreators = creators.filter(
+          (creator) =>
+            (creator.isApproved || isAdmin) &&
+            (!status || creator.isApproved === (status === "true")) &&
+            (!genre || creator.genre.toLowerCase() === genre.toLowerCase()) &&
```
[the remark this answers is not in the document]
aaggarwal10: I think this is outdated. I alr fixed it all. Can you check again?

PR #194, opened 2022-11-24
Add route for approving creators.

CalderWhite:
## Notion ticket link

[Approve Creators Route](https://www.notion.so/uwblueprintexecs/Approve-Creators-Route-cbd6c5cc12a94da3ae64e644a47e80ca)

## Implementation description
* Check if admin
* If admin approve by id
* used RESTful URL (`/approve/:id`)

## Steps to test
1. Try to approve a creator without admin
2. Try to approve a creator with admin
3. Check database if the user is an approved creator

* Whether or not these changes propagate to the database

## Checklist

### DIFF
--- a/backend/typescript/services/implementations/creatorService.ts
+++ b/backend/typescript/services/implementations/creatorService.ts
@@ -63,11 +63,11 @@ class CreatorService implements ICreatorService {
     try {
       await Creator.update(
         {
-          isApproved: true
+          isApproved: true,
         },
         {
           where: { id: userId },
-        }
+        },
       );
     } catch (error) {
       Logger.error(
--- a/backend/typescript/services/implementations/creatorService.ts
+++ b/backend/typescript/services/implementations/creatorService.ts
@@ -61,8 +61,14 @@ class CreatorService implements ICreatorService {
 
   async approveCreator(userId: string): Promise<void> {
     try {
-      const creator = await this.getCreatorById(userId);
-      creator.isApproved = true;
+      await Creator.update(
+        {
+          isApproved: true
+        },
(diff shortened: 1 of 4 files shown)


backend/typescript/services/implementations/creatorService.ts
```diff
@@ -58,6 +58,19 @@ class CreatorService implements ICreatorService {
       throw error;
     }
   }
+
+  async approveCreator(userId: string): Promise<void> {
+    try {
+      const creator = await this.getCreatorById(userId);
+      creator.isApproved = true;
```
CalderWhite, line 65: This is hackey and I'm not sure if the change will propagate to the database. If there is a better way to do this (setters & getters for example) then I would opt to use that method instead.

Shrinjay: Yeah I'm pretty sure this isn't how our ORM works. Looking at the docs for Sequelize, using a model instance is probably your best bet, use the findByPk method to get the instance from ID then update + save. 

Test it with the following steps:
```
sudo docker exec -it ccbc-db-1 bash
psql -U postgres -d ccbc
select * from users;
```
make sure the users table reflects your changes

Shrinjay (changes requested):

backend/typescript/services/implementations/creatorService.ts
Shrinjay: https://sequelize.org/docs/v6/core-concepts/model-instances/

take a look at userService.ts implementation around line 229

CalderWhite:
Updated to use the orm 👍

PR #213, opened 2023-02-05
Update Creator Data Model

CalderWhite:
## Notion ticket link

[Creator Data Model Updates](https://www.notion.so/uwblueprintexecs/Creator-Data-Model-Updates-7131a5deb59940b0ae61a35e6847dc3c)

## Implementation description
* Followed ticket notes

## Steps to test
1. TODO, left comments on unclear ticket items in notion

* I think this should be compatible with the front end but to be sure we should do an E2E test

## Checklist


backend/typescript/migrations/20230213004442-update-creator-model.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+  async up(queryInterface, Sequelize) {
```
Shrinjay, line 2: So, it looks like this just generates an empty migration file. Looking further into it, sequelize doesn't automatically generate migrations... unlike django's ORM or sqlalchemy or literally anything else...

Apparently the best way to do this is to just generate the model again using the command described here: https://sequelize.org/docs/v6/other-topics/migrations/#creating-the-first-model-and-migration . Look at the types defined in the model file and used them here. This should recreate the model file.

Shrinjay (changes requested):
Everything looks solid!! Except the migrations, those are empty haha, I provided some instructions on fixing that

Shrinjay (approved):
Merging this before we get the migrations stuff figured out to unblock Khloe and some other issues

PR #218, opened 2023-03-01
Caitlin/creator presentations section

caitlinkwan:
## Notion ticket link

[Booker: Creator Presentations Section](https://www.notion.so/uwblueprintexecs/Booker-Creator-Presentations-Section-3a34b29662644d06a116f038ef51645a)

## Implementation description
* Implemented the UI for the Presentations section of the Creator Profile

Screen Shot 2023-03-01 at 1 25 53 PM
Screen Shot 2023-03-01 at 1 28 27 PM

## Steps to test
1. Go to http://localhost:3000/users/userid

Update command:
`UPDATE creator SET presentations = array['{"name":"Readings","details":"description", "age_groups":"Primary","audience_size":30,"locations":"libraries","languages":"English, French","special_equipment":"Projector for powerpoints","is_virtual”:true,”is_bringing”:true,”in_person_rate":400,"virtual_rate":300}'::json] WHERE id=1;`

## Checklist


frontend/src/components/pages/CreatorProfile/CreatorPresentations.tsx
```diff
@@ -0,0 +1,113 @@
...
+              <Flex direction="column">
+                <Text mb="16px">{pres.description}</Text>
+                <Flex>
+                  <Flex direction="column" width="30%" mr="10px">
+                    <Text textStyle="h2" fontWeight={700} mb="17px">
+                      Additional Information
+                    </Text>
+                    {additionalInfoNames.map((name, index) => {
+                      const res =
+                        pres[
```
Shrinjay, line 83: nit: Make sure to always use descriptive variable names

frontend/src/components/pages/CreatorProfile/CreatorPresentations.tsx
```diff
@@ -0,0 +1,113 @@
...
+              <Flex
+                backgroundColor="#E2E8F0"
+                borderRadius="100px"
+                height="40px"
+                width="40px"
+                alignItems="center"
+                justifyContent="center"
+                mr="16px"
+              >
+                {(pres.name === "Readings" && <BsBook />) ||
```
Shrinjay, line 56: nit: For contsants like "Readings" or "Workshops", use a variable like READINGS_NAME to store them

frontend/src/components/pages/CreatorProfile/Carousel.tsx
```diff
@@ -0,0 +1,172 @@
...
+  languages: string[];
+  avSetup: string;
+  remote: boolean;
+  saleOrSigning: boolean;
+  price: string;
+  mediaGallery: string[];
+};
+
+// SAMPLE DATA FOR TESTING ------------------------------
+export const presentationsTest: Presentations[] = [
```
Shrinjay, line 28: This should come from the server, and the correct way to add test data is by updating the seeding script and getting it from the server. Hardcoding this means we can't figure out any issues with the API until we actually call it. 

After looking into this, I did find quite a few issues with the creator presentations implementation on the backend actually, so we get to fix them as part of this. 


So, I'd recommend
- Getting rid of this and adding the presentations to the seeding data in `seeders/sample-creators-****` 
- Now, we need to fix the presentations field in the creator model. Right now, they point to another model, but that's obviously not the case since it's a JSON field. Delete `presentations.model.ts` and instead create an interface like `interface Presentation {... all the proprties of the presentation model .. }` 
- One fix you need to make is that `photos` should be an array of strings, not a string
- Now, you can run the seeder and get presentations into the database. May deal w some issues here, but then you can test with real data instead.

frontend/src/components/pages/CreatorProfile/CreatorProfile.tsx
```diff
@@ -74,6 +76,7 @@ const CreatorProfile = (): React.ReactElement => {
               </Flex>
 
               <CreatorOverview currentCreator={currentCreator} />
+              <CreatorPresentations presentations={presentationsTest} />
```
Shrinjay, line 79: Don't commit test code

frontend/src/components/pages/CreatorProfile/CreatorPresentations.tsx
```diff
@@ -0,0 +1,113 @@
...
+                            res}
+                        </Text>
+                      );
+                    })}
+                  </Flex>
+                  <Flex direction="column" width="70%">
+                    <Text textStyle="h2" fontWeight={700} mb="17px">
+                      Media Gallery
+                    </Text>
+                    <Carousel images={pres.mediaGallery} />
```
Shrinjay, line 101: Should use `pres.photos` not `pres.mediaGallery`

Shrinjay (changes requested):
Make sure to test your code using real requests for the API and test using seeding rather than hardcoding test data. This will also require you to make some changes to the creator model definition which we can pair on. Good work so far!

backend/typescript/seeders/2022.10.20T21.57.52.sample-creators.ts
```diff
@@ -57,6 +57,63 @@ const seedCreators: CreatorSeed[] = [
     availability: ["TODO"],
     book_covers: ["TODO1", "TODO2"], // array of URLs
     isReadyForReview: true,
+    presentations: [
```
NathanSouphanthong, line 60: Hey Caitlin, if you're having trouble with seeding the presentations you can try inserting a json array of presentations into a pre-existing creator using Postgres.

`docker exec -it ccbc-db-1 bash`
`psql -U postgres -d ccbc`
`UPDATE creator SET publications = array['{"title":"The Cat in the Hat","publisher":"Random House", "publication_year":"1957","notes":"One of the best-selling books of all time"}'::json] WHERE id=1;`

In your case you would, swap out publications for presentations and id for the creator you want to insert the presentations into. Let me know if you have any questions :D

Shrinjay (approved):

Shrinjay (approved):

